PR #27, opened 2021-02-16
Add support for local assets

ajgon:
By default, WUD fetches assets from various CDNs. This is an issue for people like me, who like to keep their self-hosted environment as private as possible.

This PR address that problem, by adding `WUD_LOCAL_ASSETS` environment variable, which triggers assets serving, either from CDN (default) or locally.

P.S. I'm not a JS developer, so hopefully my code is not too horrible :)


fmartinou:
Thank you very much for your contribution!
Actually, I coded the UI as simple as possible to get something very quickly.
I plan to rewrite it in more "industrial" way (vuejs or reactjs) so this issue should be solved in the end.

Though, in the meantime, I agree with you; wud should be self-sufficient at runtime and shouldn't depend on external CDNs.

I think I'll rework your PR a little bit to keep things very simple.

I'd pack the needed assets in the app (as you already did) and that's all.
I think there is no need to offer the choice to users to rely on CDN versus local assets.
Local assets will be the only supported way.

Would it be ok for you?

ajgon:
@fmartinou Sounds awesome, thank you!

fmartinou (commented):
Will be simplified to serve local assets only

fmartinou:
@ajgon
WUD can now serve local assets instead of relying on CDN in version `3.2.3`

Best regards

PR #29, opened 2021-02-17
Fix image name processing

ajgon:
Storing image name in container object data is rather odd in Docker and shouldn't be considered reliable. Consider situation from the documentation, and run the container using tag:

```
docker run -d --name wud -v "/var/run/docker.sock:/var/run/docker.sock" -p 3000:3000 fmartinou/whats-up-docker
```

When you visit the page, the image is displayed properly as `fmartinou/whats-up-docker` with `latest` version. But what would happen if you run the same image, using it's ID instead of the name?

```
docker run -d --name wud -v "/var/run/docker.sock:/var/run/docker.sock" -p 3000:3000 sha256:2709895d99806a15a9050cae18208df4a54822e69752a977a271dd9cc22dea31
```

Now you may see, that image name is displayed improperly (`sha256` was parsed as image name, and the hash as version).
Unfortunately some automated tools used to orchestrate containers rely solely on those IDs (instead of names) and this pose a problem. This patch is an attempt to address that, and changes the default behavior to use image tags instead of container image alias. Only if those tags can't be found (which shouldn't happen), container image name is used.

### DIFF
--- a/app/watchers/providers/docker/Docker.js
+++ b/app/watchers/providers/docker/Docker.js
@@ -339,7 +339,7 @@ class Docker extends Component {
         const creationDate = containerImage.data.Created;
 
         // Parse image to get registry, organization...
-        const parsedImage = parse(container.data.Image);
+        const parsedImage = parse((containerImage.data.RepoTags || [])[0] || container.data.Image);
         const version = parsedImage.tag || 'latest';
         const parsedVersion = semver.coerce(version);
         const isSemver = parsedVersion !== null && parsedVersion !== undefined;


fmartinou:
Hi @ajgon ,

Thanks for the improvement. Indeed sha256 versioning wasn't supported so far.

However I have some concerns regarding images which would be associated with multiple tags.

I did some tests using the `nginx@sha256:14536d83ca3128923ee7c2f7f4f285e023abd40f3ccdc8911f56cd4119558506` image.
(this image is also tagged as `1.19-alpine` and as `1.19.7-alpine` on the Hub)

**Test 1**
- Just run a container using the image `nginx@sha256:14536d83ca3128923ee7c2f7f4f285e023abd40f3ccdc8911f56cd4119558506`
- Result: the list of the associated tags is `[]`

**Test 2**
- pull `nginx:1.19-alpine`
- run a container using the image `nginx@sha256:14536d83ca3128923ee7c2f7f4f285e023abd40f3ccdc8911f56cd4119558506`
- Result: the list of the associated tags is `["1.19-alpine"]`

**Test 3**
- pull `nginx:1.19-alpine`
- pull `nginx:1.19.7-alpine`
- run a container using the image `nginx@sha256:14536d83ca3128923ee7c2f7f4f285e023abd40f3ccdc8911f56cd4119558506`
- Result: the list of the associated tags is `["1.19-alpine", "1.19.7-alpine"]`

**Test 4**
- pull `nginx:1.19.7-alpine`
- pull `nginx:1.19-alpine`
- run a container using the image `nginx@sha256:14536d83ca3128923ee7c2f7f4f285e023abd40f3ccdc8911f56cd4119558506`
- Result: the list of the associated tags is `["1.19-alpine", "1.19.7-alpine"]`

**Test 5**
- pull `nginx:1.19-alpine`
- pull `nginx:1.19.7-alpine`
- run a container using the image `nginx:1.19-alpine`
- Result: the list of the associated tags is `["1.19-alpine", "1.19.7-alpine"]`

It seems that depending on which tags have or not been pulled, side effects can occur.

What do you think?

ajgon:
Yeah, tag management in docker sucks, and there is no reliable way to always determine them correctly. That's why I included fallback to `container.data.Image` as a failsafe.

I don't have good solution for that - in rare edge-cases, we may end up without tags - maybe ask hub for them then? (not sure if it's even possible though).

fmartinou:
Apparently [Integration tests are broken](https://travis-ci.org/github/fmartinou/whats-up-docker/builds/759407946#L541) so I guess there are regressions though.

Maybe would it be safer if we can detect upstream if the image reference is a tag or a sha256?
Then:
- if the reference is a tag, we parse the Tag name (`container.data.Image`)
- if it's a sha256, we try to get the first repo tag (`(containerImage.data.RepoTags || [])[0]`)

I'll give it a try.

fmartinou:
It should be working now.

Before I perform the release, may you test the [latest](https://hub.docker.com/layers/fmartinou/whats-up-docker/latest/images/sha256-1cca41bd613ae71051980d443751e315efa2b009db466f9542ba07755f36619b?context=explore
) tag of wud on your end to check if everything is working as you're expecting?

ajgon:
@fmartinou Tested it locally and it's looking good! 👍 Later today I'll check it on my server, and let you know if I hit some edge-cases :)

Thanks!

ajgon:
So, I deployed it to my machine and all the images are displayed properly now - thanks again! 🤘

fmartinou:
Great!
I won't be available during this weekend so I'll perform the release next week (likely next Tuesday).

fmartinou:
It's available in version `3.3.0`.

PR #32, opened 2021-03-24
Prometheus metrics with update availability of the images

ajgon:
This is a simple metrics, which allows display a simple counter in Grafana, stating how many images need to be updated. We can't use `wud_images` for that, because it can return multiple records for the same image.


fmartinou:
Hi,
Thanks for the PR!

You wrote:
`We can't use wud_images for that, because it can return multiple records for the same image`.

Images recorded in the store are normally [unique for the combination](https://github.com/fmartinou/whats-up-docker/blob/master/app/store/index.js#L78):
`watcher+registryUrl+image+version+includeTags+excludeTags`

So you shoudn't have more than 1 record per image unless there is something buggy?

Can you explain further?
Maybe is the root issue there?

ajgon:
Hey, sorry for late reply - got busy weekend :)

Hmmm... maybe I'm doing something wrong in my query then? Can you provide me an example query, which returns number of images available for update? Because my problem is, when I do the aggregation on `wud_images`, I aslo get previous version updates included. So for example if I updated postgres image twice before, I'm getting postgres counted as two images available for update.

fmartinou:
Hi,
Sorry for my late answer as well... ;)

I performed a test with 3 different versions of nginx at the same time.
If I want to get the detail per version, I can use the query
```
sum by(registry_url, image, os, architecture, version, new_version, updated) (wud_images)
```

which gives the following table on Grafana



If I want to have the sum per image (regardless of the versions), I can use the query
```
sum by(registry_url, image, os, architecture, new_version, updated) (wud_images)
```
(In fact, I just remove the tag `version`)

which gives me the following table on Grafana


Would it solve your issue?

fmartinou:
And to get the total number of images to be updated (regardless of the versions), you could wrap your queries into a `count`:
```
count(sum by(registry_url, image, os, architecture) (wud_images))
```

ajgon:
Definitely I'm doing something wrong. For testing, I made a setup where two of fifty images have new versions. 
The first query: `sum by(registry_url, image, os, architecture, version, new_version, updated) (wud_images)` returns all images I have at the moment (regardless of them being updateable or not). 
Second query: `sum by(registry_url, image, os, architecture, new_version, updated) (wud_images)` does the same.
The count query returns `50` - which is the number of my images.

Expected behaviour is 2 - either as a table listing those two images which needs update or as a number.

Maybe I'm misconfiguring something in Grafana? Should I do anything else, except of setting Visualization to table, query format to table and query instant to true?

P.S. When I do the last (count) query in prometheus this is the result which I get:

2021-04-07-144606_3572x648_scrot

fmartinou:
Hi,

You're right; something were missing :/

I did some modifications:
- I added a new `to_be_updated` tag on the `wud_image` prometheus metrics
- I reset the Prometheus gauge on every scrap request to get rid of old images
([See details here](https://github.com/fmartinou/whats-up-docker/commit/4c66e2f780f30037339a96de89c5c072b2c5d0dc))

So now you could be able to filter on images to be updated only.

Examples:

Without the `to_be_updated` filter
We can see that both nginx versions are listed (including the one which is already up to date)
without_filter

With the `to_be_updated` filter
We can see that only the nginx version to be updated is listed
with_filter

Could you give a try to the `fmartinou/whats-up-docker:develop` version and tell me if it solves your issue?

ajgon:
Works like a charm! I believe, resetting prometheus gauge, was the main thing which resolved this issue, but I wouldn't mind `to_be_updated` flag as well :) 

If I notice any quirks with current behavior, I'll reopen this issue, but for now this change resolves my issue.

Thank you!

fmartinou:
It's now integrated in the `3.4.0` release.

PR #141, opened 2022-06-02
fixed authentik Redirect URIs/Origins

depuits:


fmartinou:
Hi @depuits 
Thank you for having fixed the typing mistake!

PR #146, opened 2022-07-19
Fix missing semver transforms

jasoncodes:
This PR adds missing `transformTag` calls to fix the following semver tag errors when using the `wud.tag.transform` label:

* `tag.semver` in the container result being `false` when the transformed tag should be recognised as semver. This seems to result in updates not working at all as it falls back to digest mode on a fixed tag.
* `updateKind.semverDiff` in the container result being `unknown` instead of `major`/`minor`/`patch`/etc.
* Processing failing with `Cannot destructure property 'major' of 'versionSemver' as it is null.` when `wud.link.template` is configured.

### DIFF
--- a/app/model/container.js
+++ b/app/model/container.js
@@ -194,8 +194,8 @@ function addUpdateKindProperty(container) {
                         const isSemver = container.image.tag.semver;
                         if (isSemver) {
                             const semverDiff = diffSemver(
-                                container.image.tag.value,
-                                container.result.tag,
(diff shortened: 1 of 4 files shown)


fmartinou (approved):
Thank you for your **high-quality** contribution!
(I mean the code, the effort to add unit tests and so on...)

LGTM 👍 

I temporary merge your work to the `fix/semver_transforms` branch.
As soon as Travis build it successfully, I release it.

PR #279, opened 2023-05-05
Fix IFTTT default event value in docs

matthewdias:


fmartinou:
Hi @matthewdias ,

Thanks for the fix 👍 .

PR #284, opened 2023-05-09
Fix watcher variable in triggers doc

jebabin:
The variable "watcher" listed for _SIMPLEBODY and _SIMPLETITLE miss ${}


fmartinou:
Thank you :)

PR #301, opened 2023-07-13
Add a trigger to make docker-compose changes to a git repo

patmagauran:
This new trigger will clone a specified git repo in the container and make any version changes to the docker compose file there. After updating the file, it will commit and push the changes. 

As a small side feature, I also added a variable for the docker-compose trigger to disable recreating the containers and only update the compose. I set that to be forcefully true for this trigger as it would not make sense to update the repo and recreate the container, but not update the actual local files. 

This would be useful for a flow of updating the git repo immediately and then letting another script auto sync at a specified maintenance time. This maintains the concept that the compose file should be the ultimate source of truth for the containers.


patmagauran:
I am leaving in draft state until I get this working within the docker container, rather than in the development mode.

patmagauran:
I have successfully run this locally and in a docker environment. I currently am running it in my home docker setup.

patmagauran:
Found issue when we have multiple instances of the same repo. Need to fix

Issue #89, opened 2021-12-03
Publishing all docker information to MQTT


mcrodrig:
Hi,

I would like to have all available dockers information in mqtt regardless whether there is an update for it or not. This is possible when I enable HASS (true value), but not when it is disable (false value).

Thanks,

fmartinou:
Now available in the latest release `6.5.0`

PR #402, opened 2024-06-17
Proposal for the inclusion of additional mqtt items

dkadioglu:
Proposal for the inclusion of additional items (total_count, update_count, update_status, running) in regular MQTT trigger, which before that were only included when HASS integration is enabled. The approach also adds an additional config parameter to enable or disable publishing to the discovery topics usually used within HASS. The documentation has also been changed to cover this change.

This approach works for me. However, I am not sure, if you accept this "kinda hacky" approach. If you want to discuss it, please ask.

Just for the completeness: This would be a fix for https://github.com/fmartinou/whats-up-docker/issues/89


app/triggers/providers/mqtt/Hass.js
```diff
@@ -124,44 +128,56 @@ class Hass {
...
-        await this.publishDiscoveryMessage({
-            discoveryTopic: watcherUpdateStatusDiscoveryTopic,
-            stateTopic: watcherUpdateStatusTopic,
-            name: `Watcher ${container.watcher} container update status`,
-            options: {
-                payload_on: true.toString(),
-                payload_off: false.toString(),
-            },
-        });
+        if (this.configuration.hass.discovery) {
```
fmartinou, line 131: Why not wrapping all the ` this.publishDiscoveryMessage` calls into a single `if (this.configuration.hass.discovery) {}`?

dkadioglu: Is [this](https://github.com/fmartinou/whats-up-docker/pull/402/commits/94ca029b822f4b54d9fc6fe5990ebb9388e49f34) what you had in mind?

dkadioglu: It would be nice, if this could be merged. Could you review the modification, please?

fmartinou: Thanks for the change regarding the `if` parts👍 

I'm just noticing one last thing:

With your PR, all users who are currently using `HASS` will have to update their conf before upgrading because `WUD_TRIGGER_MQTT{trigger_name}_HASS_DISCOVERY` will now be by default `false` regardless the value of `WUD_TRIGGER_MQTT_{trigger_name}_HASS_ENABLED`

Do you see a use case where someone would like to integrate to HASS but without enabling the automatic discovery?

I'm wondering if we should not automatically toggle `WUD_TRIGGER_MQTT{trigger_name}_HASS_DISCOVERY` to `true` when `WUD_TRIGGER_MQTT_{trigger_name}_HASS_ENABLED` is enabled?

It should be doable to implement that by changing the Joi validation API directly on the line:
```
discovery: this.joi.boolean().default(false),
```
(changing the `false` by an evaluation of the sibling property `hass.enabled` 🤔 )

What do you think about?

dkadioglu: You're right, breaking the configuration for existing users should be avoided. Your proposal makes absolutely sense. However, as you maybe can see, I have problems to find the right implementation. Could you help me and tell me exactly how I can evaluate the sibling as you describe?

fmartinou: I've not tested by myself but the code is using the [Joi library](https://joi.dev/) and Joi seems to support this use case.

I'd say by combining `when` (to make the conditional work) and `ref` (to reference another property) functions.
- https://joi.dev/api/?v=17.13.3#anywhencondition-options
- https://joi.dev/api/?v=17.13.3#refkey-options

I'll try to find some time to test that if you don't succeed finding the right implementation.

And once again; thank you for your contribution!

dkadioglu: Thank you very much for giving me a push into the right direction. As far as I could test it, it seems to work now as expected: When I only set `WUD_TRIGGER_MQTT_FHEM_HASS_ENABLED=true`, discovery messages are published. When I set `WUD_TRIGGER_MQTT_FHEM_HASS_DISCOVERY=false`, discovery messages are not published. What do you think?

fmartinou:
Thats seems perfect; thank you very much!

I merge the PR :+1: 

Then I need a little time to fix the CI and publish a new release version.
(it should be ok by the end of the week)

fmartinou:
Now available in the latest release `6.5.0`

PR #408, opened 2024-06-24
feat: save update-available in url on value change

royto:
This PR allows to save the `update-available` parameters in URL when clicking the `Update available` switch

image

image


fmartinou:
Thank you very much for your contribution :+1

PR #413, opened 2024-07-07
fix: typos in Prometheus docs

jburnham:
The Prometheus label is `update_available` so this fixes the example query which can be used in a dashboard panel.


fmartinou:
Thank you very much 👍

PR #445, opened 2024-10-03
Add warning when filtering all tags.

sebbi08:
Feel free to change the warning message


fmartinou:
Thanks for the contribution 👍 

I'll just slighty change your code to avoid "non semver" tags to cause this warn message whereas an empty array of candidate tags is a normal situation.

Issue #210, opened 2023-01-28
[Feature Request] 'Update' button


boomam:
For those of us that dont want to automate updating, it would be good to retain watching ability, but have the ability to press an 'update' button to update.

thedafa:
I would like automatic update for selected containers (a list of free-to-update items could be given to wud with variables/labels), but for other very important ones I would like to be more in control. So an manual update button would be an awesome idea.

PiotrEsse:
+1

ducadk:
+1

Zatow:
I'm the same with @boomam, I want to monitor my containers for updates and deploy them with an "Update" button

derekoharrow:
This would be awesome. I'm currently using Watchtower to auto-update some containers, and WUD to tell me when the rest need updating. Having this option to manually update certain containers would mean I could just use WUD.
(I guess I could also run two instances of WUD - one just to notify and one to auto-update, but that seems excessive).

JvdMaat:
This has also been discussed as part of the update entity in HA. Issue is closed, but referencing it here to maintain easy-access history: https://github.com/fmartinou/whats-up-docker/issues/186#issuecomment-1344603845
There was some initial discussion on how to get this setup.

kirk56k:
I would love to see this feature. I'll add my plus one to this as well.

fmartinou:
I've started to work on a proof-of-concept few weeks ago...
It involves a big rework and because I don't have much time these days, it will last a bit before it's done.
But it will be ready sooner or later 😁 .

loeffelpan:
What about Hass Integration? You create update-entities in hass. What does clicking "Update" in hass do today? And what should it do concerning this feature?

fmartinou:
> You create update-entities in hass. What does clicking "Update" in hass do today? 

Currently you cannot click on the hass `update` button because it has been disabled.
You can only `skip` the available update

![image](https://github.com/fmartinou/whats-up-docker/assets/16061231/c5ce9f68-c65a-4fb1-8bfa-661ceb1d3b45)


> What about Hass Integration? And what should it do concerning this feature?

This github issue will provide `triggers` the capability to receive container update commands (`mqtt` topic, `http` request, `telegram` command, `ui` action...) and `wud` will update the container using the appropriate `watcher` (`docker`, `docker-compose`...)

Regarding Home-Assistant integration:
- All containers managed by WUD will be added as Home-assistant `devices`
[...]

PR #474, opened 2024-10-28
Adds install feature to HTTP trigger

nopoz:
Adds install feature to HTTP trigger. The intention of the feature is to allow for manual container upgrades when an update is available. This requires a listening webhook external to WUD that will respond to the HTTP call and trigger the upgrade process.  

When "WUD_TRIGGER_HTTP_{trigger_name}_INSTALL" is set to `true`:
1.  An "Install" button appears next to the upgrade version in the UI. See below for an example screenshot. When there is no upgrade available the install button will not show either. 
2. Clicking the "Install" button executes the HTTP trigger.
3. If "INSTALL" env variable is true, normal scheduled watch triggers do not happen for this trigger. 
4. Only one INSTALL env variable is allowed across all triggers. If more than one is set, the UI throws an error message indicating that only one is allowed and the HTTP trigger is not executed. 
5. Adds "install" key to the containers API endpoint for the UI to action on. 
6. Data sent by the HTTP trigger to the destination web hook will contain `actionType: install`. This could be used by the webhook to filter received HTTP calls if necessary. 

With the addition of the install key to the API it should be possible to add the INSTALL env feature to other trigger types as well. 

Should resolve (at least partially) #210

Screenshot:
install_button_example

### DIFF
--- a/app/api/container.js
+++ b/app/api/container.js
@@ -2,11 +2,13 @@ const express = require('express');
 const nocache = require('nocache');
 const storeContainer = require('../store/container');
 const registry = require('../registry');
-const { getServerConfiguration } = require('../configuration');
+const { getServerConfiguration, getTriggerConfigurations } = require('../configuration');
+const HttpTrigger = require('../triggers/providers/http/Http');
 
 const router = express.Router();
 
 const serverConfiguration = getServerConfiguration();
+const triggerConfigurations = getTriggerConfigurations();
 
 /**
  * Return registered watchers.
@@ -32,7 +34,41 @@ function getContainersFromStore(query) {
  */
 function getContainers(req, res) {
     const { query } = req;
-    res.status(200).json(getContainersFromStore(query));
+    const containers = getContainersFromStore(query);
+
+    // Access the 'install' flag from the Http trigger configuration
+    const httpTriggers = triggerConfigurations && triggerConfigurations.http;
+
+    console.log('Trigger Configurations:', JSON.stringify(triggerConfigurations, null, 2));
+    console.log('HTTP Triggers:', JSON.stringify(httpTriggers, null, 2));
+
+    let installEnabled = false;
+
+    if (httpTriggers) {
+        // Collect all trigger names that have install enabled
+        const triggersWithInstall = Object.keys(httpTriggers).filter((triggerName) => {
+            const triggerConfig = httpTriggers[triggerName];
+            const installValue = String(triggerConfig.install).toLowerCase();
+            return installValue === 'true';
+        });
+
+        if (triggersWithInstall.length === 1) {
+            // Only one trigger has install enabled
+            installEnabled = true;
+        } else if (triggersWithInstall.length > 1) {
+            // More than one trigger has install enabled
+            installEnabled = 'multiple';
+            console.warn('Multiple triggers have install enabled; install action will be disabled.');
+        }
+        // If no triggers have install enabled, installEnabled remains false
+    }
+
+    const containersWithInstallFlag = containers.map((container) => ({
+        ...container,
+        install: installEnabled,
+    }));
+
+    res.status(200).json(containersWithInstallFlag);
 }
 
 /**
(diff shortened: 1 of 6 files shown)


fmartinou:
Hi,

Thanks for the proposal 👍 

I'm wondering if it coudn't be implemented in a more generic way 🤔 ?

I'm thinking about something like:

- Adding a common option to all triggers to trigger either `manually` or `automatically` (the current behavior)
- If the trigger is configured to be `manually` triggered, then an action button is displayed in the UI to trigger it

Some benefits I can see:
- no differences between triggers reporting updates and triggers really updating containers
- no need to control if multiple manual triggers are attached to the container; just display one button for each of them and let the user choose 😄 

Another question; when a user clicks the "install" button, how does he get feedback about if the container has been updated with success or not?

_ps; I would rather named the button `Update` instead of `Install` to better fit in the current terminology used around_

Once again; thank you very much for your great work; let me know what you think!

nopoz:
>     * Adding a common option to all triggers to trigger either `manually` or `automatically` (the current behavior)

I'm not sure if an update/manual option for all triggers makes sense as many are for notification purposes only? 

The ones I have in mind for manual trigger of updates are:
- HTTP (this one)
- Docker
- Docker Compose
- Script (Current working on this one - for executing a local script mounted inside the countainer via a docker volume)

The rest are mostly for notifications, right? I wouldn't anticipate many more install methods other than that. When I'm done with the new Script trigger, I'll add a comment linking to my Script branch so you can see how HTTP and Script are integrated together. 

> no need to control if multiple manual triggers are attached to the container; just display one button for each of them and let the user choose 😄

This could get messy - like how would users identify which button is for which install in the UI? Is there a practical use case for having more than one Update method? We could go for one as it is and if there is a bunch of users clamoring for more than one then it can be re-evaluated then? 

> Another question; when a user clicks the "install" button, how does he get feedback about if the container has been updated with success or not? 

This is dependent on the users external webhook and it's a variable I don't think we can completely account for. The code does check for error/success for the HTTP call but this is just the signal to the webhook. Presumably the true upgrade success response is going to be seen by the webhook service itself and not WUD. However, if the update was truly successful, a refresh of the containers page should show that there is no longer an update available for that particular container. 

> _ps; I would rather named the button `Update` instead of `Install` to better fit in the current terminology used around_

This had crossed my mind as well as to what would be the best name for the button. I'm fine with changing it to "Update", makes sense to me.

nopoz:
For reference, here is the link for the script_trigger branch which builds on top of the http_install branch referenced here. 

https://github.com/nopoz/wud/tree/script_trigger

fmartinou:
> The ones I have in mind for manual trigger of updates are:
- HTTP (this one)
- Docker
- Docker Compose
- Script (Current working on this one - for executing a local script mounted inside the container via a docker volume)

I'd add these one to the list of candidates for manual updates:
- IFTTT
- KAFKA
- MQTT

But generally speaking, why not allowing to manually trigger all configured triggers from the UI? 🤔 
It would allow users to more easily test their integrations.
Currently they have to restart `wud` and perform a global `watch` which is consuming from their Docker Hub API quota.

> This could get messy - like how would users identify which button is for which install in the UI? Is there a practical use case for having more than one Update method? We could go for one as it is and if there is a bunch of users clamoring for more than one then it can be re-evaluated then?

Putting user-friendly names on the triggers from the configuration would be an easy solution to display user-friendly buttons 😁 .
Most users will only have one configured; I agree.
But it will be easier to implement and it will fit all cases.

We could imagine users who want to get "automatic" notifications when an update is available (e.g. by mail, slack...); there will be a button for it in the UI (not used very often; mainly for setup/testing purpose)
And at the same time, there will be another button to "manually" trigger the container update (e.g. by running a script).

> This is dependent on the users external webhook and it's a variable I don't think we can completely account for. The code does check for error/success for the HTTP call but this is just the signal to the webhook. Presumably the true upgrade success response is going to be seen by the webhook service itself and not WUD. However, if the update was truly successful, a refresh of the containers page should show that there is no longer an update available for that particular container and it's current version will have been bumped to the previous update version.

Totally agree; most updates will occur asynchronously.
`wud` also have the possibility to listen to docker events so a simple refresh of the UI should do the trick.
In the future, adding websocket for UI realtime updates could improve that part a little bit...

nopoz:
Having the update function for all triggers sounds okay. I suppose it's up to the user if they want to clutter up their UI or not.  :smile:

I've added some new commits for the Script trigger type in my script_trigger branch - https://github.com/nopoz/wud/tree/script_trigger

This branch has a containers page refresh mechanism integrated with the containers endpoint API. After the update is completed, and the old container id is pruned from the store, the API `/api/containers/${this.container.id}/install` value is updated which instructs the containers UI to refresh itself so the newly updated container shows up.

fmartinou:
Hi,

I just added the following features:
- The support of a new `command` trigger allowing to run arbitrary commands/scripts
- The support or running any trigger from the UI/API
- The support of `auto` / `manual` triggers to allow users to decide if they want to update automatically or not

It's maybe not the exact idea of yours but it should solve the same issue 🤞 .

PR #523, opened 2024-12-23
Fix digest accept headers

### LINKED ISSUE #522: LSCR error fetching digest
In Registry.js, when running get `getImageManifestDigest(image, digest)` with a digest, it fails for lscr.io with an error saying: `OCI manifest found, but Accept header does not support OCI`.

It can be fixed by adding two more MIME-types to the accept header.
```js
const responseManifests = await this.callRegistry({
    image,
    url: `${image.registry.url}/${image.name}/manifests/${tagOrDigest        }`,
    headers: {
        Accept: 'application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.index.v1+json, application/vnd.docker.distribution.manifest.v2+json, application/vnd.oci.image.manifest.v1+json',
    },
});

mlofjard:
Fixes #522

### DIFF
--- a/app/registries/Registry.js
+++ b/app/registries/Registry.js
@@ -128,7 +128,7 @@ class Registry extends Component {
             image,
             url: `${image.registry.url}/${image.name}/manifests/${tagOrDigest}`,
             headers: {
-                Accept: 'application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.index.v1+json',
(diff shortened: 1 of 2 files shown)


mlofjard:
Fixed the tests, but the CI build timed out after 10 mins.

fmartinou:
Thank you very much 👍

PR #541, opened 2025-01-05
⭐ [UI/API] - Add UI option to group the container by their docker compose project

sebbi08:
Hi, I added the docker compose project name to the container info and made it, so the containers are groupable by their compose project.

### DIFF
--- a/app/prometheus/container.js
+++ b/app/prometheus/container.js
@@ -12,7 +12,14 @@ function populateGauge() {
     gaugeContainer.reset();
     storeContainer.getContainers().forEach((container) => {
         try {
-            gaugeContainer.set(flatten(container), 1);
+            const flatContainer = flatten(container);
+            const flatContainerWithoutLaels = Object.keys(flatContainer)
+                .filter((key) => !key.startsWith('labels_'))
+                .reduce((obj, key) => {
+                    obj[key] = flatContainer[key];
+                    return obj;
+                }, {});
+            gaugeContainer.set(flatContainerWithoutLaels, 1);
         } catch (e) {
             log.warn(
                 `${container.id} - Error when adding container to the metrics (${e.message})`,
@@ -71,7 +78,6 @@ function init() {
             'update_kind_remote_value',
             'update_kind_semver_diff',
             'error_message',
-            'compose_project',
         ],
     });
     log.debug('Start container metrics interval');
--- a/app/prometheus/container.js
+++ b/app/prometheus/container.js
@@ -71,6 +71,7 @@ function init() {
             'update_kind_remote_value',
             'update_kind_semver_diff',
             'error_message',
+            'compose_project',
         ],
     });
     log.debug('Start container metrics interval');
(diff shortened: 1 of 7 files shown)


app/watchers/providers/docker/Docker.js
```diff
@@ -746,6 +746,7 @@ class Docker extends Component {
                 variant,
                 created,
             },
+            composeProject: container.Labels['com.docker.compose.project'],
```
sebbi08, line 749: One problem I currently see here is that this will only be added to new containers that are not yet in the store

sebbi08:
Here also a preview of what it looks like

image

fmartinou:
Hi,


Thank you for your contribution; this is a great idea! 👍 

> One problem I currently see here is that this will only be added to new containers that are not yet in the store

That's not a big deal; users can easily accept to wipe their store.

In addition, if the feature is released with the upcoming `v8` version, the store will be automatically wiped because of other non backward compatible changes 😇 .

---

I have, though, one interrogation regarding the design of the feature:

Rather than capturing only the `com.docker.compose.project`, maybe we can capture **all** container's labels?
That would allow, at the same time, to display them in the UI 🤔.

And if so, maybe the "Group by compose" toggle could be replaced by a "Group by" selection list where all labels can be selected to group containers on?

In that way, the feature will be more "generic" and not "docker compose" specific only?

What do you think?

app/prometheus/container.js
```diff
@@ -12,7 +12,14 @@ function populateGauge() {
     gaugeContainer.reset();
     storeContainer.getContainers().forEach((container) => {
         try {
-            gaugeContainer.set(flatten(container), 1);
+            const flatContainer = flatten(container);
+            const flatContainerWithoutLaels = Object.keys(flatContainer)
```
fmartinou, line 16: Little typo there `flatContainerWithoutLaels` -> `flatContainerWithoutLabels`

fmartinou:
That's very nice; thank you very much 👍 

I'll find some time in the next days to test it.

Regarding the large numbers of container labels, the autocomplete combobox should be enough for now 👌

fmartinou:
I tested it and it looks very nice! 👍 

Regarding the UI, I think that repeating the label name next to the value would improve the readability?

```vue
    <div v-if="this.groupingLabel && this.previousContainer?.labels?.[this.groupingLabel] !== this.container.labels?.[this.groupingLabel]">
      <div class="text-h6">
        {{ this.groupingLabel }} = 
        {{ this.container.labels?.[this.groupingLabel] ?? "(empty)" }}
      </div>
      <v-divider class="pb-3"></v-divider>
    </div>
```

Examples:
image

image

fmartinou:
I merge the code as it, and I'll fix my minor remarks by myself.

Thank you very much for your contribution! 👍

hugo-id:
I see that this is closed, but is there chance we can get these groups to foldable/collapsible?

I group by project, but then I can collapse these new dynamic groups, so I can focus only the things I'm trying to work on?

Issue #590, opened 2025-01-30
Feature Request: Multiple SemVer tags for Docker images


RoboMagus:
It'd be nice if WUD would have multiple SemVer tags per release, just like many other Docker images offer (e.g. Traefik, Mosquitto, HomeAssistant, etc.). 

This allows users to auto update to versions they are comfortable with (mostly useful for patch releases) while still being safe from unintended auto updates to newer (major) versions.

The reasoning behind this is best explained [here](https://medium.com/@mccode/using-semantic-versioning-for-docker-image-tags-dfde8be06699) (Scroll down to Why This Matters).

PR #591, opened 2025-01-30
Add Major and Minor version release tags to docker build & push

RoboMagus:
Implements #590.

This PR adds additional tags when publishing new docker images. E.g. when publishing new image `4.5.6`, the tags for this image will be `latest`, `4`, `4.5`, and `4.5.6`.

I'm open to suggestions on different implementations for this feature. I usually deal with Github Workflows instead of Travis so there may be more elegant ways of implementing this for the existing release flow that I am unaware of.


fmartinou:
Hi,

That's a very good idea; thank you!

It looks good to me; I merge it as it is.

PR #648, opened 2025-03-28
add possibility to sort by oldest

jlssmt:
Old container are a security risk
and by wud.tag.include it's possible to not get notified about updates, if regex is wrong, or author of the image changes the format.

So I implemented the possibility to sort by oldest images:
image

Dates are only shown if filter is set.


fmartinou:
Hi,

Thank you very much for your contribution.
I added some minor fixes: https://github.com/getwud/wud/pull/671

The feature will be available in the next release.

Issue #647, opened 2025-03-27
[Question] Major with latest tag


jlssmt:
Hello,

is it possible to detect major/minor when latest tag is used?

This is a major update -> v1.11.0 -> v2.32.0
![Image](https://github.com/user-attachments/assets/c6f8b87d-a2d4-4f52-ab80-2e64f1401ffb)
Label `'wud.tag.include=^v\d+\.\d+\.\d+$$'` is set

Tecnically it should be possible to get via digest the current installed semver.

If not I would definitely need the docker compose trigger.
But I I'm also watching remote sockets.
Do I need an extra trigger for every single docker-compose.yaml? If yes all compose trigger will be shown for every container even if the compose don't belongs to the container, right? 

Thanks in advance!

thecoshman:
Wouldn't it make more sense to use a `v1` tag rather than `latest`? And perhaps combine that with the the regex being `^v1\.\d+\.\d+$`

fmartinou:
Hi @jlssmt,

Thank you for your question!

It is not possible to determine whether an update is `major`, `minor`, or `patch` when tracking the `latest` tag. The Docker registry manifest for `latest` only provides an image digest (hash), without any inherent semantic version metadata.

To get `major`/`minor`/`patch` detection and threshold triggers, the container must be tagged with a SemVer version tag (e.g., `image: myapp:1.2.3`), with a tag inclusion filter like `wud.tag.include=^\\d+\\.\\d+\\.\\d+$$`.

Closing this question as it reflects standard Docker tag design.

PR #653, opened 2025-03-30
semver of container with latest tag can now sometimes extracted

jlssmt:
i started the work to get the semver of running containers when tag latest is used

#647


jlssmt:
functionality should work now.
documentation will be added after first review

open question: watcher env to defaultly try to get local semver from org.opencontainers.image.version?
-> leads to problems in some containers.
f.e. ollama misuses the label and writes the ubuntu version inside (20.04)
docker run -d --rm --name ollama ollama/ollama
docker inspect ollama

jlssmt:
> > org.opencontainers.image.version
> 
> Important to note: This is an open standard and does not necirarily have any reflection to "tags" or versioning employed in tags.
> 
> Its important to note, because it can lead to false positives.

For me this was the only way to receive the tag version from latest.

jlssmt:
> I also see a number of changes that are not fundemental to this PR, might be worth it to cut it up in logical smaller PRs

I left WUD some time ago to renovate.
Please feel free to make adjustments or close the PR.

I never received any feedback here, and JavaScript without types is really a mess for contributors.

glebaccon:
Sounds like a great idea to me. I hope the merge will happen.

scttbnsn:
This has been integrated into [WUD-CE](https://github.com/CodesWhat/whatsupdocker-ce) as of 9.0.0-ce. Thanks for the contribution!

Issue #450, opened 2024-10-10
Not showing containers from local machine


kajvans:
I am now a few days into fixing my install since it failed after i created a symlink to a new install location for docker because the drive was full from 50gb of docker images.

i tried two approches:
```
      - WUD_WATCHER_Nas_HOST=192.168.1.225
      - WUD_WATCHER_Nas_PORT=2375
      - WUD_WATCHER_Nas_WATCHALL=true
```

and i have tried with creating a docker volume and setting the socket to the volume.

neither of these ways has worked.
When i use curl from a remote machine i get all the data so it is something inside the container failing to grab it.
The logs dont show anything really special only that it tries to search for some of the containers but it does not show them.

the logs:
```
18:27:38.230  INFO whats-up-docker/watcher.docker.nas: Register with configuration {"host":"192.168.1.225","watchall":true,"port":2375,"socket":"/var/run/docker.sock","cron":"0 * * * *","watchbydefault":true,"watchevents":true}
18:27:38.231  INFO whats-up-docker/watcher.docker.nas: Cron scheduled (0 * * * *)
18:27:38.245  INFO whats-up-docker/watcher.docker.raspberry: Register with configuration {"host":"192.168.1.175","watchall":true,"port":2375,"socket":"/var/run/docker.sock","cron":"0 * * * *","watchbydefault":true,"watchevents":true}
18:27:38.245  INFO whats-up-docker/watcher.docker.raspberry: Cron scheduled (0 * * * *)
18:27:38.247  INFO whats-up-docker/registry: No authentication configured => Allow anonymous access
18:27:38.249  INFO whats-up-docker/authentication.anonymous.anonymous: Register with configuration {}
18:27:38.255  WARN whats-up-docker: Anonymous authentication is enabled; please make sure that the app is not exposed to unsecure networks
18:27:38.261  INFO whats-up-docker/api: Server listening on port 3000 (HTTP)
18:27:39.247  INFO whats-up-docker/watcher.docker.nas: Cron started (0 * * * *)
18:27:39.268  INFO whats-up-docker/watcher.docker.nas: Listening to docker events
18:27:39.273  INFO whats-up-docker/watcher.docker.raspberry: Cron started (0 * * * *)
18:27:39.275  INFO whats-up-docker/watcher.docker.raspberry: Listening to docker events
18:27:39.566  WARN whats-up-docker: nas_speedtest-tracker - No Registry Provider found
[...]

PR #657, opened 2025-04-08
Handle image lookup failures to prevent the `No such image...` error along with a blank container list

d0gkiller87:
Catch image lookup failures to prevent one/more unknown images from throwing the `No such image...` error while being queried, therefore presenting the user with a blank container list ([#450](https://github.com/getwud/wud/issues/450))


d0gkiller87:
tagging another related issue https://github.com/getwud/wud/issues/400

Issue #630, opened 2025-03-11
WUD does not handle docker API code 409 / removal of container already in progress well


kquinsland:
I am putting together a PoC using WUD as a replacement for watchtower.

I have a super simple "hello, world" flask app that I'll call `demo` here.
I can run it like so:

```shell
docker run --rm --name=demo --label wud.watch=true <...>
```

> [!NOTE]
> the `--rm` flag is there to clean up the container when it stops but it'll make life difficult for WUD...

Then, I start up WUD like so:

```shell
docker run --rm --name wud \
  -v "/var/run/docker.sock:/var/run/docker.sock" \
  -p 3000:3000 \
  -e "TZ=America/Los_Angeles" \
  -e "WUD_WATCHER_LOCAL_SOCKET=/var/run/docker.sock" \
  -e "WUD_WATCHER_LOCAL_CRON=*/1 * * * *" \
  -e "WUD_REGISTRY_CUSTOM<...> \
  -e "WUD_LOG_LEVEL=debug" \
  -e WUD_WATCHER_LOCAL_WATCHBYDEFAULT="false" \
  -e WUD_TRIGGER_DOCKER_DEMO_AUTO=true \
  ghcr.io/getwud/wud:latest
```

This should check for new images every minute and, if it finds one, pull it down and re-create the container, right?
So I can then trigger the CI/CD process to build a new version of the container, `demo`.

Shortly after the CI/CD pipe pushes the new container image, WUD correctly realizes that there's a new image and faithfully begins pulling it down.

```shell
15:19:20.510  INFO whats-up-docker/watcher.docker.local: Cron finished (1 containers watched, 0 errors, 1 available updates)
15:19:20.514  INFO whats-up-docker/trigger.docker.demo: Pull image test-registry.local/test/demo:poc (container=local_demo-test)
15:19:24.474  INFO whats-up-docker/trigger.docker.demo: Image test-registry.local/test/demo:poc pulled with success (container=local_demo-test)
15:19:24.474  INFO whats-up-docker/trigger.docker.demo: Stop container demo-test with id 95<...>7b (container=local_demo-test)
15:19:25.024  INFO whats-up-docker/trigger.docker.demo: Container demo-test with id 95<...>7b stopped with success (container=local_demo-test)
15:19:25.024  INFO whats-up-docker/trigger.docker.demo: Remove container demo-test with id 95<...>7b (container=local_demo-test)
[...]

PR #681, opened 2025-05-03
Add docker trigger support for containers that are set to auto remove

Tom-V:
Fixes docker trigger to support containers that are set to auto remove.
Now in that case, instead of removing the container after stopping it, it will wait until the container is removed.

Fixes: #630


fmartinou:
Awesome! Thank you very much 🙏

Issue #636, opened 2025-03-21
[QUESTION] How do you define a valid email address for the SMTP trigger?


mhosse:
Hi,

I configured a SMTP trigger which was ignored because the email address (root@foo.lan) was considered as invalid. Unfortunately I did not find any information what makes an email address a valid one for WUD. Imo root@foo.lan is fine. But of course, it's not using an official registered domain name. That's not needed in my case, because it's only used locally.

mhosse:
Imo I can answer my own question: Looking at the source code in https://github.com/getwud/wud/blob/25bea0cf1220a343b42b4c6a9cdef761c7cd507c/app/triggers/providers/smtp/Smtp.js#L21 and checking https://joi.dev/api/?v=17.13.3#stringemailoptions the default is to check for registered TLD. But "lan" isn't in that list.

Is there any reason not to change that?

Tom-V:
I've created a pull request to make this possible.

fmartinou:
Fixed

PR #696, opened 2025-06-07
Add support for custom TLDs in SMTP trigger

Tom-V:
Adds a new `allowcustomtld` boolean flag to the SMTP trigger config, allowing support for email addresses with custom (non-standard) TLDs like `.lan`, `.local` or `.internal`. When set to `true`, domain validation will no longer restrict to known public TLDs.

Default behavior remains unchanged. Includes tests for both standard and custom TLD cases

Requested by #636


fmartinou:
Thanks!

PR #699, opened 2025-06-17
Fix prerelease variable not populating in link template generation

### LINKED ISSUE #584: [Bug] `prerelease` label not working on v8
The `prerelease` in `wud.link.template` seems broken on WUD v8, it appears as `undefined` on the http link when it's valid.
I've tested the same scenario on WUD v7.2.0, and it works as expected.

Scenario:
```
      - 'wud.watch=true'
      - 'wud.tag.include=^\d+\.\d+\.\d+\.\d+-\w+$$'
      - 'wud.tag.transform=^(\d+\.\d+\.\d+)\.(\d+)-\w+$$ => $$1-$$2'
      - 'wud.link.template=http://localhost/$${major}.$${minor}.$${patch}.$${prerelease}'
```

Input:
```
1.2.3.4-foobar
```

Result:
```
http://localhost/1.2.3.undefined
```

As per the docs [1], this configuration would be still ok on v8 right?

[1] https://getwud.github.io/wud/#/configuration/watchers/?id=transform-the-tags-before-performing-the-analysis

SisyphusMD:
Previously, the assignment for the prerelease variable was referencing its own (uninitialized) value, i.e., prerelease = prerelease[0], rather than the parsed semver array. As a result, prerelease was always set to an empty string instead of the intended prerelease identifier.

This commit updates the code to assign prerelease from versionSemver.prerelease[0], ensuring the variable pulls the correct value from the parsed semver object.

Fixes #584

### DIFF
--- a/app/model/container.js
+++ b/app/model/container.js
@@ -100,7 +100,7 @@ function getLink(container, originalTagValue) {
         patch = versionSemver.patch;
         prerelease =
             versionSemver.prerelease && versionSemver.prerelease.length > 0
-                ? prerelease[0]
+                ? versionSemver.prerelease[0]
                 : '';
     }
     return eval('`' + container.linkTemplate + '`');


fmartinou:
Thank you very much for your contribution! 👍

PR #706, opened 2025-06-26
Configure 60s default jitter in docker watcher

gregj-d:
Hi 👋 

This is a follow up to https://github.com/getwud/wud/issues/684, I also work on Docker hub and this change updates to the latest `node-cron` version that includes an option for jitter, and sets a default jitter of 1 minute in the Docker watcher, but is configurable via the `WUD_WATCHER_LOCAL_JITTER` environment variable.

This will help alleviate the spikes we see at the very top of every hour by spreading requests out over the first minute.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -30,7 +30,7 @@
                 "lokijs": "1.5.12",
                 "mqtt": "5.10.3",
                 "nocache": "4.0.0",
-                "node-cron": "3.0.3",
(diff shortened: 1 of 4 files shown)


fmartinou:
Hi!

I didn't know that my project was causing noise on the Hub 😂 😇 .
I'm happy to merge your PR and make it available asap.

fmartinou:
Fix available in the latest `8.1.0` version.

PR #717, opened 2025-07-19
Added LSCR icon

webysther:
All [distribution](https://github.com/simple-icons/simple-icons/issues/13635) is based in CNI and for unknown I think using linuxcontainers will the fine.

Allow in future this to have custom icon like I [explain here](https://github.com/getwud/wud/discussions/716) will be nice.


fmartinou:
Sounds great; thanks!

Issue #726, opened 2025-07-31
'Not Authorized' on image pull when using private ECR


scadorel:
Hello,

When using WUD with a private ECR registry, newer version are properly found by the app but pulling the image is failing with error "(HTTP code 401) unexpected - unauthorized: Not Authorized".

While debugging/investigating the code, I found out that the credentials given to the pull command is using AWS access key id and secret access key instead of the autorization token used for API calls. I'll forward a PR soon with the fix :)

WUD version: 8.1.1 (latest at the time of writing).

PR #727, opened 2025-07-31
Fix authentication for private ECR registry

scadorel:
Fixes issue #726.
This is my first PR on this repo, let me know if I need to change anything 👍

### DIFF
--- a/app/registries/BaseRegistry.js
+++ b/app/registries/BaseRegistry.js
@@ -58,7 +58,7 @@ class BaseRegistry extends Registry {
     /**
      * Common auth pull credentials
      */
-    getAuthPull() {
+    async getAuthPull() {
         if (this.configuration.login && this.configuration.password) {
             return {
                 username: this.configuration.login,
--- a/app/registries/Registry.js
+++ b/app/registries/Registry.js
@@ -321,7 +321,7 @@ class Registry extends Component {
      * @returns {}
      */
 
-    getAuthPull() {
+    async getAuthPull() {
         return undefined;
     }
 }
--- a/app/registries/providers/acr/Acr.js
+++ b/app/registries/providers/acr/Acr.js
@@ -53,7 +53,7 @@ class Acr extends Registry {
         return requestOptionsWithAuth;
     }
 
-    getAuthPull() {
+    async getAuthPull() {
         return {
             username: this.configuration.clientid,
             password: this.configuration.clientsecret,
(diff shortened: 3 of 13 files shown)


thecoshman (commented):
Changes look sensible to me, though as I'm not that familiar with js dev, what's the reason so much ended up needing to be marked as `async` here?

scadorel:
Thanks @thecoshman and sorry for the late reply.
For ECR, I need to call some AWS SDK methods to generate a token, which requires an await (so the getAuthPull() needed to be moved to async as well). But getAuthPull is from the parent Registry class used in various places, thus the update to async in all usage.

The alternative (what I did for a workaround at first), was to keep the generated token from authenticate() in Ecr.js inside a member property, and reuse it in the getPullAuth(). This was simpler to implement as only code changes in Ecr.js but, from my point of view, not really clean as it forces an hidden dependency between authenticate() and getPullAuth() methods, which is not really good pratices. If someone calls getPullAuth() before authenticate() in the future, it will fail; so that's why I moved to async in getPullAuth().

I'm not a javascript expert, but from what I've saw on various sources on internet, that's apparently the best practices for such cases. I would have prefered to avoid changing so much code but that's probably cleaner than my first workaround :)

Code from workaround 1:

```js
    async authenticate(image, requestOptions) {
            // [...]
            const authorizationToken = await ecr.getAuthorizationToken().promise();
            const tokenValue = authorizationToken.authorizationData[0].authorizationToken;
            requestOptionsWithAuth.headers.Authorization = `Basic ${tokenValue}`;

            // CODE ADDED
            const auth = Buffer.from(tokenValue, 'base64').toString().split(':');
            this.pullAuth = {
                username: auth[0],
                password: auth[1],
            };
            // [...]

        return requestOptionsWithAuth;
    }

    // Before
    getAuthPull() {
        return this.configuration.accesskeyid
            ? {
                  username: this.configuration.accesskeyid,
                  password: this.configuration.secretaccesskey,
              }
            : undefined;
    }

    // After
    getAuthPull() {
        return this.configuration.accesskeyid ? this.pullAuth : undefined;
    }
```

thecoshman:
@scadorel oh that makes perfect sense. Thanks for the clear explanation - certainly helps me get to grips a bit more with js dev... I've been trying to avoid it, but it's caught up with me :sweat_smile:

fmartinou:
Hi, looks good, thank you.
May you resolve the merge conflicts before I merge, please?

JTufarelli-BAH:
Hello, are we able to merge this in? I would love this feature because I have the same issue with ECR as well.

scadorel:
Hello @fmartinou, branch rebased and updated with latest changes from main.

JTufarelli-BAH:
Hey, just to give you some feedback. I cloned your fork @fmartinou and built the Docker container to test it out. Everything works perfectly! Once this is merged in I can start using the public build of WUD again.

PR #783, opened 2025-10-12
Document password colon limitation in basic auth

### LINKED ISSUE #782: Basic auth passwords don't support special characters
WUD's basic authentication refuses to log in users with (some?) special characters in their password.

The password that didn't work for me was: ```LN<=:.9oF`H,T)g=ALy/BowM3|z;0]```

It worked perfectly after I changed to a password containing only characters from `a-zA-Z0-9`.

Maximilian-Nesslauer:
## Summary
- Add warning about passwords containing colon characters (`:`) not being supported
- References issue #782

## Context
Basic authentication fails when passwords contain colon characters due to a known bug in the `passport-http` library. The library incorrectly parses the Authorization header by splitting on all colons instead of only the first one, which truncates passwords containing `:`.

This documentation update warns users about this limitation until the underlying library issue is resolved.

## Changes
- Added a warning box in the basic auth documentation explaining the colon limitation
- Advises users to avoid colons in passwords as a workaround

Fixes #782

### DIFF
--- a/docs/configuration/authentications/basic/README.md
+++ b/docs/configuration/authentications/basic/README.md
@@ -20,6 +20,8 @@ or use single quotes in Bash commands \
 or escape them in Bash commands \
 `WUD_AUTH_BASIC_JOHN_HASH="\$apr1\$aefKbZEa$ZSA5Y3zv9vDQOxr283NGx/"`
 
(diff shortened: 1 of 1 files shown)


fmartinou:
Thanks for your contribution :)

PR #787, opened 2025-10-17
Fix Ntfy trigger configuration masking

david-uhlig:
When attempting to mask sensitive `auth` parameters in the `maskConfiguration()` method, they would incorrectly get masked with `undefined`, instead of a `e****e` form. This happens due to the method accessing incorrect properties in the configuration object. The PR fixes this.

This test fails without the changes:

```js
// Ntfy.test.js
test('maskConfiguration should mask sensitive data', () => {
    ntfy.configuration = {
        auth: {
            user: 'user',
            password: 'password',
            token: 'token',
        },
    };
    expect(ntfy.maskConfiguration()).toEqual({
        auth: {
            user: 'u**r',
            password: 'p******d',
            token: 't***n',
        },
    });
});
```

### DIFF
--- a/app/triggers/providers/ntfy/Ntfy.js
+++ b/app/triggers/providers/ntfy/Ntfy.js
@@ -36,9 +36,9 @@ class Ntfy extends Trigger {
             ...this.configuration,
             auth: this.configuration.auth
                 ? {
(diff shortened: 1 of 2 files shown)


fmartinou:
Thank you!

Issue #774, opened 2025-10-04
Home Assistant 2025.10 MQTT deprecated object_id


Vruschok:
Home Assistant 2025.10 deprecated object_id and suggests to use default_entity_id in MQTT entity configurations.

It logs
> The configuration for entity update.wud_container_host_xxx uses the deprecated option `object_id` to set the default entity id. Replace the `"object_id": "wud_container_host_xxx"` option with `"default_entity_id": "update.wud_container_host_xxx"` in your published discovery configuration to fix this issue, or contact the maintainer of the integration that published this config to fix this. This will stop working in Home Assistant Core 2026.4

scttbnsn:
This has been fixed in [WUD-CE](https://github.com/CodesWhat/whatsupdocker-ce) as of 9.0.0-ce — MQTT discovery now uses `default_entity_id` instead of the deprecated `object_id`, compatible with Home Assistant 2025.10+.

fmartinou:
Hello! This issue has been addressed and is available in WUD releases.

Thank you for your feedback! If you still experience any issues with the latest WUD release, please feel free to open a new issue.

PR #791, opened 2025-10-23
Home-assistant sensors: replaced object_id by default_entity_id

Vruschok:
Fixes issue #774.

Simple replacement of existing property "object_id" by "default_entity_id" in Hass.js and Hass.test.js.

The tests run as before.


Vruschok:
Does not work by a simple replacement. Default_entity_id has to include domain: "sensor.my_sensor".

PR #800, opened 2025-11-01
Add support for major-only and minor-only thresholds

thelpfy:
Added additional '{semver}-only' trigger thresholds.
Also fixed a typo.

I had some use cases where I wanted most of my containers to be auto-updated on minor and patch versions, but I still wanted to review major updates. As I did not want to get spammed everyday by new update notification which I do not care about as they are treated automatically, I needed some major-only and minor-only trigger thrsholds.


thelpfy:
With this we also need to update documentation etc. but let me know if this is something you will merge or why this was not included initially :)

thelpfy:
I have just noticed that the branch that was created from within the GitHub UI is named 'patch-1' :/
If this is an issue, I can create a cloned pull-request with a better name, for example: 'feature/advanced_semver_thresholds' or something similar

fmartinou:
Hi, this makes perfectly sense 👌 
Thank you for your contribution

Issue #275, opened 2023-04-28
WUD container is crashing


attilabalazsy:
I'm having issues with WUD as it continuesly crashing after couple of days/hours. The crash is mostly related to some communication or parsing issues, so most probably because of an incorrect error handling. 

WUD is monitoring 3 docker instances, one is running locally and 2 are accessed through TCP.
WUD is running on RPI4 (ARMv8)

This is the stack trace from the last crash:
23:00:00.742  INFO whats-up-docker/watcher.docker.rpi4: Cron started (0 * * * *)
23:00:13.565  INFO whats-up-docker/watcher.docker.rpi2: Cron finished (6 containers watched, 0 errors, 1 available updates)
23:00:15.464  INFO whats-up-docker/watcher.docker.rpi4: Cron finished (12 containers watched, 0 errors, 3 available updates)
23:00:15.633  WARN whats-up-docker/watcher.docker.docker02: Error when processing (502 - "<html><body><h1>502 Bad Gateway</h1>\nThe server returned an invalid or incomplete response.\n</body></html>\n\n") (container=docker02_Container-health-check)
23:00:15.637  WARN whats-up-docker/watcher.docker.docker02: Error when processing some containers (Cannot read properties of undefined (reading 'tag'))
23:00:15.637  INFO whats-up-docker/watcher.docker.docker02: Cron finished (0 containers watched, 0 errors, 0 available updates)
23:15:53.534  INFO whats-up-docker/watcher.docker.docker02: Status changed from running to exited (container=docker02_ntfy)
23:15:57.605  INFO whats-up-docker/watcher.docker.docker02: Status changed from running to exited (container=docker02_uptime-kuma)
23:15:57.790  INFO whats-up-docker/watcher.docker.docker02: Status changed from running to exited (container=docker02_apprise)
undefined:1
[...]

Issue #793, opened 2025-10-26
[Bug] WUD keeps crashing with JSON syntax error


holzerseb:
I'm running Docker Desktop with many disks mounted. I presume the number of mounts causes the issue. Error is a little redacted, but generally, the issue occurs if:

- Docker Engine restarts (boot or manual engine restart)
- After some containers get re-created

If WUD is full re-created (docker compose down && up), the issue goes away until any of above events occur.

Here's the log:

```
10:04:01.202  INFO whats-up-docker/api: Server listening on port -0000 (HTTP)
10:04:02.188  INFO whats-up-docker/watcher.docker.somethingsomething: Listening to docker events
<anonymous_script>:1
[...]

Issue #792, opened 2025-10-24
WUD crashes on a Javascript error


thesnowdog:
While doing my initial setup and adding "wud.watch=true" label to each of my containers, the WUD container kept crashing with the following error. 

18:59:54.420  INFO whats-up-docker: WUD is starting (version = 8.1.1)
18:59:54.423  INFO whats-up-docker/store: Load store from (/store/wud.json)
18:59:54.426  INFO whats-up-docker/store: Create Collection app
18:59:54.428  INFO whats-up-docker/store: Migrate data from version undefined to version 8.1.1
18:59:54.430  INFO whats-up-docker/store: Create Collection containers
18:59:54.430  INFO whats-up-docker/prometheus: Init Prometheus module
18:59:54.449  INFO whats-up-docker/registry: Register all components of kind trigger for provider ntfy
18:59:55.016  INFO whats-up-docker/trigger.ntfy.private: Register with configuration {"mode":"batch","topic":"dockerNotify","batchtitle":"${containers.length} updates available","simplebody":"App ${container.name} can be updated on ${watcher}","auth":{},"url":"https://****.org","simpletitle":"${container.name} on ${watcher}","auto":true,"threshold":"all","once":true}
18:59:55.016  INFO whats-up-docker/trigger.ntfy.private: Registering for auto execution
18:59:55.019  INFO whats-up-docker/registry: Register all components of kind registry for provider ecr
18:59:55.223  INFO whats-up-docker/registry.ecr.public: Register with configuration {}
18:59:55.223  INFO whats-up-docker/registry: Register all components of kind registry for provider gcr
18:59:55.228  INFO whats-up-docker/registry.gcr.public: Register with configuration {}
18:59:55.228  INFO whats-up-docker/registry: Register all components of kind registry for provider ghcr
18:59:55.231  INFO whats-up-docker/registry.ghcr.public: Register with configuration {}
18:59:55.231  INFO whats-up-docker/registry: Register all components of kind registry for provider hub
18:59:55.238  INFO whats-up-docker/registry.hub.public: Register with configuration {"password":"d**********************************U","login":"t***"}
18:59:55.239  INFO whats-up-docker/registry: Register all components of kind registry for provider quay
18:59:55.243  INFO whats-up-docker/registry.quay.public: Register with configuration {}
18:59:55.243  INFO whats-up-docker/registry: Register all components of kind registry for provider lscr
[...]

Issue #675, opened 2025-04-25
Simple fix for "SyntaxError: Unterminated string in JSON"


LukeBurke99:
Running into a major issue with using a remote docker instance that causes the entire container to crash. This bug has been mentioned previously and after some research it looks like someone has already found a fix: https://github.com/getwud/wud/issues/559

I haven't done a pull request because I'm not too sure on how to do it and not sure if one was already created for this issue but I would really appreciate a fix on this, as it stops my container every day.

urbaman:
Hi,

I stumbled upon the same problem watching for events through a docker proxy.

I'm just restarting a container:

```
10:09:50.107  INFO whats-up-docker/watcher.docker.docker3: Listening to docker events
10:09:50.119  INFO whats-up-docker/watcher.docker.docker2: Listening to docker events
10:09:50.120  INFO whats-up-docker/watcher.docker.docker1: Listening to docker events
<anonymous_script>:1
{"status":"die","id":"e2b01d676d8d456af41e23b83119f9283ca14f4db7ea75398959b352ed43f30f","from":"docker.io/valkey/valkey","Type":"container","Action":"die","Actor":{"ID":"e2b01d676d8d456af41e23b83119f9283ca14f4db7ea75398959b352ed43f30f","Attributes":{"com.docker.compose.config-hash":"3006335ac643bf158b9ae5cd13cca5a35d2668859ea188c9e6adfd51d68f2776","com.docker.compose.container-numb

SyntaxError: Unterminated string in JSON at position 384 (line 1 column 385)
    at JSON.parse (<anonymous>)
    at Docker.onDockerEvent (/home/node/app/watchers/providers/docker/Docker.js:383:34)
    at IncomingMessage.<anonymous> (/home/node/app/watchers/providers/docker/Docker.js:372:51)
    at IncomingMessage.emit (node:events:507:28)
    at IncomingMessage.emit (node:domain:489:12)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
    at Readable.push (node:internal/streams/readable:390:5)
    at HTTPParser.parserOnBody (node:_http_common:129:24)
    at Socket.socketOnData (node:_http_client:558:22)
    at Socket.emit (node:events:507:28)
    at Socket.emit (node:domain:489:12)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
[...]

PR #867, opened 2025-12-03
fix(docker-watcher): process events only when buffer is complete

khborgnis:
This is a pull request implementing the code in [this comment](https://github.com/getwud/wud/issues/559#issuecomment-2661457342) by @sirinsidiator :

> The problem is that in https://github.com/getwud/wud/blob/main/app/watchers/providers/docker/Docker.js#L369 the streamed chunks are processed directly on arrival, which ends up attempting to parse a partial docker event.
> I modified the file locally and with this change it works for me
> I'd also add a try catch to the JSON parse in onDockerEvent, as it seems to crash the whole container whenever there is an error

My addition is adding a couple of tests to verify the functionality.

These are probably dupes of the oldest issue: #601, #675, #792, and #793.

Fixes: #275

### DIFF
--- a/app/watchers/providers/docker/Docker.js
+++ b/app/watchers/providers/docker/Docker.js
@@ -445,7 +445,16 @@ class Docker extends Component {
                     this.log.debug(err);
                 }
             } else {
-                stream.on('data', (chunk) => this.onDockerEvent(chunk));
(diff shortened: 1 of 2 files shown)


fmartinou:
Thank you :)

PR #871, opened 2025-12-09
feat: Allow disabling Prometheus metrics via env var

### LINKED ISSUE #766: WUD idle CPU usage?
What is WUD doing when it's idle?
Looking at the CPU usage I see that it's at 0.25% baseline with ~5% spikes every 5 seconds:
Image
Shouldn't it be completely still when it's not actively checking the images?
Could it be the /health endpoint called by docker healthcheck?
Or maybe it's trying to look for watchers, even if `WUD_WATCHER_{watcher_name}_WATCHEVENTS=false`?

difagume:
## Description
This PR resolves #766 by addressing the report of idle CPU usage spikes.
To give users more control and resource efficiency, this change introduces a new environment variable `WUD_PROMETHEUS_ENABLED`.
- **Default behavior**: `true` (Prometheus enabled).
- **Opt-out**: Users can set `WUD_PROMETHEUS_ENABLED=false` to completely disable the Prometheus module and its background metrics collection.

### DIFF
--- a/app/configuration/index.js
+++ b/app/configuration/index.js
@@ -150,6 +150,25 @@ function getServerConfiguration() {
     return configurationToValidate.value;
 }
 
+/**
+ * Get Prometheus configurations.
+ */
+function getPrometheusConfiguration() {
+    const configurationFromEnv = get('wud.prometheus', wudEnvVars);
+    const configurationSchema = joi.object().keys({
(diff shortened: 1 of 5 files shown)


fmartinou:
Thanks for your contribution; LGTM 👌

PR #874, opened 2025-12-20
Fix issues from recent PRs

avargaskun:
A mix of PRs that were merged recently had the following issues:

PR #840
- Code is throwing errors instead of returning the filtered list of tags (this change is effectively reverted by this PR)

PR #841 
- Code is attempting to match regex against boolean field: `container.image.tag.semver`

PR #842 
-  Code calling `isDigestToWatch` should pass parsedImage but instead is passing the domain which causes a NPE
-  The label `wud.watch.digest` should be respected regardless of registry (it's up to the user to deal with rate-limiting)

Adding unit test coverage to protect against regressions in these areas.


avargaskun:
Closing this PR as the fixes contained within have been included with #893

PR #879, opened 2025-12-28
Fix(docker-compose-trigger): YAML Parsing Error "Excessive alias count indicates a resource exhaustion attack"

DevonLobb:
The [YAML library](https://eemeli.org/yaml/#tojs-options) has a default limit of 100 aliases, resulting in an error when the docker-compose trigger runs on a file with > 100 aliases:
`Error when parsing the docker-compose yaml file (Excessive alias count indicates a resource exhaustion attack)`

This increases the alias limit to 10,000.

Similar discussion/solution in another project:
(https://github.com/portainer/portainer/issues/6211)


fmartinou:
Thank you :)

PR #892, opened 2026-01-09
Adding UI build step into Docker build process

steilerDev:
I was trying to get docker-compose trigger working and noticed, that those changes have not yet been released to DockerHub - therefore I needed to build the container locally, when I noticed that the UI is currently not build as one of the build stages (but requires a local build).

Therefore I quickly added an intermediate stage that also builds the UI in Docker to simplify the build and release process.

I've also optimized the overall Dockerfile - let me know if you think I should revert any of those changes:
- Moved metadata from base into production stage
- Removed unnecessary apk dependencies (`jq` and `git`) - I guess `tzdata` and maybe even `openssl` could be removed as well - but I don't know the app's codebase to have high confidence in this
- Removed UI sources from `.dockerignore` to enable UI build in Docker container

P.S.: It would be great if you could release a new version - 8.1.1 is 6 months old and I really need the changes to the dockercompose trigger :) Thanks!!!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,23 @@
 # Common Stage
 FROM node:24-alpine AS base
+WORKDIR /home/node/app
+
+# App dependency stage
+FROM base AS app-dependencies
+COPY app/package* ./
+RUN npm ci --omit=dev --omit=optional --no-audit --no-fund --no-update-notifier
+
+# UI stage - building UI assets
+FROM base AS ui-dependencies
+COPY ./ui ./
+RUN pwd && tree .
+RUN npm ci --no-audit --no-fund --no-update-notifier && npm run build
+
+# Release stage
+FROM base AS release 
 
 LABEL maintainer="fmartinou"
 EXPOSE 3000
-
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -12,36 +26,22 @@ ENV WUD_VERSION=$WUD_VERSION
...
+# Setup directory structure
 RUN mkdir /store
 
 # Add useful stuff
-RUN apk add --no-cache tzdata openssl curl git jq bash
-
-# Dependencies stage
-FROM base AS dependencies
+# RUN apk add --no-cache tzdata openssl curl git jq bash
+RUN apk add --no-cache tzdata openssl curl bash
```
fmartinou: git and jq were there for those who use the `cmd` trigger and could need them.

scttbnsn:
This has been integrated into [WUD-CE](https://github.com/CodesWhat/whatsupdocker-ce) as of 9.0.0-ce — our Dockerfile has multi-stage builds for both app and UI. Thanks for the contribution!

fmartinou:
Closed in favor of #1076

PR #899, opened 2026-01-15
Adding Codeberg to the default registries

### LINKED ISSUE #336: Add Codeberg to standard registries
Hi,
it would be nice if we can have Codeberg on the default registries or to configure it because if anyone host an git server with forgejo eventually you need to check the Codeberg registry about new versions of forgejo. At the moment you can configure it with the forgejo registry but if you host by your self an forgejo instance you cant setup not both Codeberg and Forgejo, therefore it would be nice if we can have an separatly configuration for Codeberg.

regner:
Codeberg has become big enough that it should be included in the default public registries supported by WUD.

Codeberg runs Forgejo and is the main/official deployment of Forgejo. In theory the Forgejo provider could have been used and just supplied the Codeberg URL, but I felt that had two issues:
- The config for the default registries is no longer consistent
- The name of the Codeberg registry within the UI would be inconsistent

So while this does add yet another provider, it's just an extension of FOrgejo, which is itself an extension of Gitea. So I think this is ok.

Closes #336


regner:
Note that I don't ever do Node development and don't have Node installed so haven't run any tests. I will try and do that later, but it's a fairly straight forward change.

fmartinou:
Thank you for your contribution

PR #935, opened 2026-02-16
Switch Grafana dashboard to use result_link

regner:
The Grafana dashboard currently uses the value of "link" to create a link to the release notes. That however links to the release notes of the currently installed version. Switching from link to result_link changes it so the generated link in Grafana takes you to the release notes of the version you would be updating to. If the container is on the newest version then link and result_link are the same.


fmartinou:
Thanks for the fix!

PR #937, opened 2026-02-16
Fixed ntfy implementation

kevinsnijder:


kevinsnijder:
Has already been fixed in recent release

PR #940, opened 2026-02-17
bugfix: Rename 'data' to 'body' in Apprise trigger

adrianmusante:
This pull request fixes an issue introduced in version `8.2.0` where the Apprise trigger stopped working.  
The request payload was being sent using the `data` field instead of `body`, which causes Apprise to return a *Bad Request* with the error:

```
Payload lacks minimum requirements
```

According to the Apprise API specification, the JSON payload must be sent in the `body` field.  
This PR updates the Apprise trigger implementation to restore the expected behavior.

For additional context, the related issue is documented here: https://github.com/getwud/wud/issues/939


fmartinou:
Thank you!

PR #946, opened 2026-02-18
fix(dockercompose): deal with services without images

sim6:
When docker-compose.yml have a service without image like use `build` instead of `image` the following error is raised and trigger do nothing.

Log error:
```
WARN whats-up-docker/container: Error when running trigger (type=dockercompose, name=mydockercompose) (Cannot read properties of undefined (reading 'includes'))
```


sim6:
I rebased it onto `main` to avoid merge conflicts.

fmartinou:
Thank you!

sim6:
:)

PR #956, opened 2026-02-23
fix: Passport auth, Playwright e2e tests, and Dockerfile UI build improvements

avargaskun:
## Summary

This PR addresses three independent improvements:

### 1. Fix improper use of Passport.js authentication middleware

The previous implementation called `passport.authenticate()` directly as route middleware on every request, which caused issues because Passport's `authenticate()` method is designed for login endpoints — it does not respect already-authenticated sessions. This meant that even authenticated users would be re-challenged on every request.

The fix introduces a `requireAuthentication` middleware that first checks `req.isAuthenticated()` (i.e., whether a session already has a verified user), and only falls back to `passport.authenticate()` if the request is not already authenticated. This is the correct pattern for session-based auth with Passport.

- Added `requireAuthentication` middleware in `app/api/auth.ts`
- Replaced direct `passport.authenticate()` calls in both `app/api/api.ts` and `app/api/auth.ts` with the new middleware

### 2. Add Playwright end-to-end integration tests for the WebUI

Adds a new `ui-e2e/` directory containing a full Playwright test suite that exercises the WUD web interface against a live running instance:

- **`tests/home.spec.ts`**: Verifies the home/dashboard page loads correctly
- **`tests/login.spec.ts`**: Tests login flow including invalid credentials and successful authentication
- **`tests/containers.spec.ts`**: Exercises the containers view — listing, filtering, and interacting with container entries
- **`tests/navigation.spec.ts`**: Validates sidebar navigation, routing between sections (Containers, Watchers, Registries, Triggers, etc.), and configuration pages

Also adds `scripts/run-ui-tests.sh`, an orchestration script that:
1. Cleans up and sets up test Docker containers
2. Builds and starts WUD
3. Waits for the service to be responsive
4. Installs dependencies and Playwright browsers
5. Runs the full Playwright test suite

### 3. Move UI build process into Dockerfile (multi-stage build)

Previously the UI was built outside the Docker image and the pre-built `ui/dist/` was copied in. This made the build dependent on a local `npm run build` step and required committing or pre-generating the dist artifacts.

The Dockerfile now includes a dedicated `ui-build` stage that:
- Copies `ui/package*.json` and installs dependencies
- Copies the full UI source and runs `npm run build`
- The resulting `ui/dist/` is then copied into the release stage via `COPY --from=ui-build`

The `.dockerignore` was updated accordingly — instead of excluding individual UI source directories, it now simply excludes `ui/dist/` (since the build now happens inside Docker).

## Test plan

- [ ] Verify authentication works for unauthenticated requests (should challenge) and authenticated sessions (should pass through without re-challenging)
- [ ] Run `scripts/run-ui-tests.sh` with a local WUD instance and confirm all Playwright tests pass
- [ ] Build the Docker image with `docker build .` and confirm the UI is correctly built and served without needing a pre-built `ui/dist/`


fmartinou:
Thank you very much! 🙏

PR #959, opened 2026-02-24
Request bearer token to authenticate with gitea/forgejo/codeberg

lysanntranvouez:
This fixes getwud/wud#797 by allowing anonymous API calls even with the v2 of the API.

I copied and slightly adjusted the existing approach and tests from Hub, which does the exact same thing.

I have tested it against real registries:
* anonymously against the public code.forgejo.org and codeberg.org registries
* with auth against a self hosted forgejo instance (v11.0.10)
* **no** gitea registry

### DIFF
--- a/app/registries/providers/gitea/Gitea.test.ts
+++ b/app/registries/providers/gitea/Gitea.test.ts
@@ -1,5 +1,8 @@
 // @ts-nocheck
+import axios from 'axios';
 import Gitea from './Gitea';
+
+jest.mock('axios');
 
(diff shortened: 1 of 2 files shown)


fmartinou:
Thank you for your contribution 👍 .
(I'll test it on a separate branch before integrating into the main branch.)

fmartinou:
I added the public code.forgejo.org registry as default to wud
=> https://github.com/getwud/wud/pull/969

PR #966, opened 2026-02-25
fix(dockercompose): handle multiple containers with same image

jarylc:
2026-02-26-clip

When multiple containers share the same image (e.g., multiple redis containers), the dockercompose trigger would fail to update all of them.

The issue was that after processing the first container and updating the compose file, subsequent containers with the same image could no longer find their service (since the image version had changed).

This fix tracks which services have already been processed to avoid duplicate updates while still notifying all containers.

Fixes issue where only one container would upgrade when multiple containers use the same image.


fmartinou:
Thanks!

Issue #970, opened 2026-02-25
Issue with prefix check


didierm-perso:
If the prefix includes digits, the current regexp is considering the digits as the first part of the semver and the real part of the semver is ignored.

<img width="643" height="393" alt="image" src="https://github.com/user-attachments/assets/cd45e483-7e60-43e4-b190-cd2b7a70a102" />

WUD is saying that we are using semver but is saying that their is an upgrade to the latest tag -> no more semver ?

If I use the tag amd64-latest, WUD is also find a semver:
<img width="643" height="393" alt="image" src="https://github.com/user-attachments/assets/b6ce0dd4-62c9-4422-be91-dbb2051398e0" />

WUD consider that the 64 is the semver.

The result is that we will check digest every time even for tags with correct semver. We can hit the rate-limit on docker hub with this kind of checks.

thecoshman:
Can you not resolve this using the include/transform labels? something like `^amd64-(.*)$` (or `$$` in the compose file) should strip of the leading amd64 part if I'm not mistaken. 
Or have I missed your point here?

didierm-perso:
Yes I'm using include/transform all the time but the PR https://github.com/getwud/wud/pull/839 is making this part automatically if you don't use include and unfortunately in case you have digits in the prefix part you have the issue I opened for.
The current code can lead to errors, hitting rate-limit on docker hub (I also tried to force the label wud.watch.digest=false and WUD is checking the digest)
For this test I used an "amd64-" prefix but it can be whatever prefix with digits.
I created a non perfect PR but I don't know how I can improve it.

fmartinou:
Hi @didierm-perso,

When image tags combine alphanumeric prefixes containing digits with a SemVer version, WUD's default tag parser needs an explicit regex filter or transform to differentiate the prefix from the semver numbers.

Using `wud.tag.include` with a precise regex or using `wud.tag.transform` allows you to isolate the numeric semver part.

Closing as expected parser behavior on non-standard prefixed tags.

didierm-perso:
Hi @fmartinou,
There is no standard for prefix tags.
You already checking for prefix with a regexp without `wud.tag.include` or `wud.tag.transform`
Why not just modify this regexp to catch a correct semver and avoid errors? (I created a PR month ago for this modification)

PR #971, opened 2026-02-25
fix: prefix result bad semver

didierm-perso:
Issue #970 

This PR is not prefect but the regexp now checks if we have at least x.y.z to start the 2nd group. This group will be used for the semver compare.
This will work with tag like amd64-1.2.3 but not with tag like amd64-1 or amd64-1.2 (those tags should use digest and not semver) If we want to keep amd64-1 or amd64-1.2 we'll need to use the include tags.

Fix PR #839

### DIFF
--- a/app/watchers/providers/docker/Docker.ts
+++ b/app/watchers/providers/docker/Docker.ts
@@ -112,7 +112,7 @@ function getTagCandidates(
         // Prefix is almost-always standardised around "must stay the same" for tags
         if (!container.includeTags) {
             const currentTag = container.image.tag.value;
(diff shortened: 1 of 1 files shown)


thecoshman:
:thinking: Not dug into the full detail of this area of the code... but isn't the problem more that this regex isn't expecting literal `.` in the potential 'semver' part of the tag? A regex a bit like `(\d+\.?)+` as the 'core' of finding semver of the top of my head feels like it'll do a better job of trying to find the semver from tags...

Also... is there no testing around this area?

PR #997, opened 2026-03-14
update OIDC library

Tom-V:
Updates OIDC library to latest version
- Allows using other claim than the default 'email' with the fallback to email if the configured one doesn't exist
- Now also supports PKCE flow instead of normal code flow
- Fixes using URL with port


fmartinou:
Thank you!

Issue #963, opened 2026-02-25
ENV option to disable digest watching


RoboMagus:
With the introduction of digest watching by default in `8.2.0` I get a long list of available updates that I simply _do not care about_, as those are covered by Watchtower.

There's the option to disable digest watching per container by adding the label `wud.watch.digest=false` to all 70+ containers that are running on the system. This should not be necessary.

Please add an option similar to `WUD_WATCHER_{watcher_name}_WATCHBYDEFAULT` which provides the default for the `wud.watch` label.
E.g. `WUD_WATCHER_{watcher_name}_WATCHDIGESTBYDEFAULT`, to provide the default for `wud.watch.digest` when the label is not specified.

Aditionally, for those that do care about digest updates there is no option in the UI to filter these out. The `update kind` dropdown can filter for `major` and `minor`, but selecting either of these will take out _both_ `patch` and `digest` updates, where users might wish to still see the `patch` updates.

fmartinou:
Hi,

> I get a long list of available updates that I simply do not care about, as those are covered by Watchtower

So shouldn't these containers simply be excluded from being watched by wud?

Either:
- Disable WUD on all containers (WUD_WATCHER_{watcher_name}_WATCHBYDEFAULT=false)
- And enable wud per container (`wud.watch=true` label)

Or 
- Enable WUD on all containers (WUD_WATCHER_{watcher_name}_WATCHBYDEFAULT=true)
- And disable wud per container (`wud.watch=false` label)

Did I misunderstand your issue?

iamtherobin:
He is asking the same I requested here (which I should have opened this ticket myself) https://github.com/getwud/wud/issues/832 This is an "ugly" change in my use case.

I already have watch by default disabled and enable wud watching for all containers I want pinned to semver so I can see the version release notes before updating. I do not need nor care to constantly be pulling images to check for digest updates that do not come with release notes. I have dockcheck setup to handle those containers for me without the need to pull images.

As mentioned it should not be necessary to go and update all the individual container labels to disable digest watching due to introducing an inconsistency. This really should have a global option to control this behavior. Three options should exist for a global switch in this case: Global disable, Global enable and Auto (Docker hub off / Other repos on). You can set auto as default if you want and just document the options and allow per container label to override the global setting.

I think that will fit to all use cases surrounding this issue.

RoboMagus:
> So shouldn't these containers simply be excluded from being watched by wud?

[...]

PR #1001, opened 2026-03-19
Add ENV option for watch digest default

RoboMagus:
See #963.

This PR provides users with the option to set the default digest watch behavior for when `wud.watch.digest` is unset.
Will only affect watch behavior when _explicitly_ set to either `true` or `false` and has 0 effect otherwise.


scr4bble:
Any chance this could be reviewed and potentially merged? :)
@fmartinou

fmartinou:
Thank you!

PR #1004, opened 2026-03-21
Use the automatic docker compose file label

pbaylas:
Rather than require the compose file to be explicitly set for each container, this PR uses the automatically set `com.docker.compose.project.config_files` if it's available. This works (at least from my testing) as long as you mount the compose files on the same path as the host in the wud container.


fmartinou:
Thanks.

Just noticed that these labels are deprecated but it's not causing any regression so... why not 😄 
(https://github.com/docker/compose/issues/10389)

PR #1023, opened 2026-03-29
Improve MQTT trigger

Endymion78:
This PR prevents the MQTT trigger creation to fail if MQTT broker is unavailable by using the synchronous MQTT client creation method instead of the asynchronous one.

This PR also publishes a new connection status sensor (and its discovery topic if enabled) which allows MQTT subscribers to know if WUD is connected to the broker. The will added to MQTT client options instructs the MQTT broker to automatically set this sensor to offline when WUD disconnects.


fmartinou:
Awesome, thank you!

PR #1038, opened 2026-04-08
Fix #819. Use manifest mediaType instead of config mediaType for single-platform manifests

### LINKED ISSUE #819: WUD does not detect updates from custom self-hosted registry (new digest ignored)
Hi,  
When using a custom self-hosted Docker registry, WUD does not detect image updates even though the digest has changed.

- The registry returns a valid manifest with a new digest (`Docker-Content-Digest` confirmed via curl).
- The container still runs an old image (checked with `docker inspect`), but WUD doesn't trigger an update.
- In the UI, the `Digest` field is missing for this container.
- With GHCR or Docker Hub, WUD detects the update and shows the digest correctly.

It seems that WUD doesn't extract or match the digest properly for custom registries.

Thanks.

micaherne:
This fixes issue #819 where digest changes from custom Docker registries were being ignored. This appears to be caused by the single-platform manifest handling selecting the wrong digest.

When a registry returns a single-platform manifest.v2+json response, the code was setting manifestMediaType to the config's media type (application/vnd.docker.container.image.v1+json) rather than the manifest's media type. This caused the code to skip the HEAD request that retrieves the correct Docker-Content-Digest header, instead returning the config digest directly.

The config digest and the manifest digest are different things - Docker's RepoDigest is a hash of the manifest, not the config blob - so the comparison against the running container's digest never matched, and digest watching never triggered.

### DIFF
--- a/app/registries/Registry.test.ts
+++ b/app/registries/Registry.test.ts
@@ -275,6 +275,55 @@ test('getImageManifestDigest should throw when no digest found', async () => {
     ).rejects.toEqual(new Error('Unexpected error; no manifest found'));
 });
 
+test('getImageManifestDigest should return manifest digest (not config digest) for single-platform application/vnd.docker.distribution.manifest.v2+json with container image config', async () => {
+    const registryMocked = new Registry();
+    registryMocked.log = log;
+    registryMocked.callRegistry = (options) => {
+        if (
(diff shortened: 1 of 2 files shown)


fmartinou:
Closed in favor of #1085

PR #1058, opened 2026-05-10
This PR introduces a new notification trigger provider for XMPP, allowing users to receive container update alerts via the XMPP protocol.

b09dan:
It implements the new `Xmpp` trigger class mimicking the structure of existing triggers and relies on the lightweight `@xmpp/client` dependency.

### Changes Included
*   **Added `@xmpp/client` dependency** to `app/package.json` to handle the underlying protocol messaging.
*   **Created `Xmpp.ts`** under `app/triggers/providers/xmpp/`, exposing a configuration schema taking `service`, `domain`, `user`, `password`, and `to` properties. 
*   **Implemented messaging logic** that properly connects, dispatches both single and batched notifications into chat stanzas, and gracefully disconnects per event.
*   **Added Unit Tests** in `Xmpp.test.ts` to validate the Joi configuration schema, data masking, and simulate `@xmpp/client` successful message dispatches. 

### Configuration Example
To use it, users will just need to set the variables such as:
```env
WUD_TRIGGER_XMPP_MYBOT_SERVICE=xmpps://chat.example.com:5223
WUD_TRIGGER_XMPP_MYBOT_USER=user
WUD_TRIGGER_XMPP_MYBOT_PASSWORD=secret
WUD_TRIGGER_XMPP_MYBOT_TO=admin@example.com

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,46 @@
+name: Docker Publish
+
+on:
+  push:
+    branches: [ "main", "feature/xmpp-trigger" ]
+    tags: [ 'v*.*.*' ]
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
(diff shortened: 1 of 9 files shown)


.github/workflows/docker-publish.yml
fmartinou, line 1: Can you delete GitHub workflow files?
As for now CI is on Travis.

app/patches/@xmpp+sasl+0.13.6.patch
fmartinou, line 1: Can you explain why that patch file?
Looks hacky; no better way?

PR #1059, opened 2026-05-12
fix(ntfy): use axios basic auth field names

antreask:
Map ntfy basic auth credentials to username/password instead of user/pass so authenticated requests work.

### DIFF
--- a/app/triggers/providers/ntfy/Ntfy.test.ts
+++ b/app/triggers/providers/ntfy/Ntfy.test.ts
@@ -118,7 +118,7 @@ test('trigger should use basic auth when configured like that', async () => {
         method: 'POST',
 
         url: 'http://xxx.com',
-        auth: { user: 'user', pass: 'pass' },
+        auth: { username: 'user', password: 'pass' },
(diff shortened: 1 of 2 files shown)


thecoshman:
@antreask I'm not a maintainer, just enthusiastic, could you expand on why this change is required? What is this actually fixing, was it ever working etc. It might help get a fix merged in faster

antreask:
@thecoshman the issue is that the ntfy trigger is using the wrong property names for Axios built-in Basic Auth option.

Axios expects this shape:

```ts
auth: {
  username: 'xxx',
  password: 'xxx'
}
```

but the ntfy trigger was passing:
```ts
auth: {
  user: 'xxx',
  pass: 'xxx'
}
```
Axios does not treat `user` and` pass` as Basic Auth credentials, so it does not generate the expected `Authorization: Basic ...` header. ntfy then receives the publish request without valid Basic Auth and replies with `401 Unauthorized`.

fmartinou:
Thanks for the fix! 👍

Issue #1052, opened 2026-05-01
Public ECR image fails with HTTP 405 even with digest watching enabled: public.ecr.aws/gravitational/teleport-distroless:18


LightMoon:
I would open it as a **bug report / feature gap**: WUD advertises ECR support, but Public ECR appears to require different handling from normal Docker Registry tag discovery. AWS documents that ECR Public supports Docker Registry HTTP API authentication, but also says the tags API is an exception; it also requires an authorization token for HTTP API requests. ([[AWS Documentation](https://docs.aws.amazon.com/AmazonECR/latest/public/public-registry-auth.html)][2])

---

## GitHub issue title

```text
Public ECR image fails with HTTP 405 even with digest watching enabled: public.ecr.aws/gravitational/teleport-distroless:18
```

## issue body

````markdown
## Summary

WUD 8.2.2 is unable to process a public Amazon ECR image from `public.ecr.aws` and repeatedly fails with HTTP 405.

The image is:

```text
public.ecr.aws/gravitational/teleport-distroless:18
````

This image is pullable and inspectable with Docker, but WUD fails to process it during every scan.

The goal is to have WUD automatically update the running container when the mutable `:18` tag changes, using digest watching.

## Environment

* WUD version: `8.2.2`
* WUD image: `getwud/wud:8.2.2`
* Registry: Amazon ECR Public
* Registry hostname: `public.ecr.aws`
* Image: `public.ecr.aws/gravitational/teleport-distroless:18`
* Container name: `teleport`
* Docker host: Ubuntu server
* Deployment method: Docker/Portainer Business stack
* WUD trigger: Docker trigger, auto execution enabled
* WUD watcher: local Docker watcher

## WUD compose configuration

```yaml
services:
  wud:
    container_name: wud
    image: getwud/wud:8.2.2
    restart: unless-stopped
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      - /opt/wud/store:/store
    environment:
      - TZ=Australia/Sydney
      - WUD_WATCHER_LOCAL_CRON=*/1 * * * *
      - WUD_WATCHER_LOCAL_WATCHBYDEFAULT=true
      - WUD_WATCHER_LOCAL_WATCHALL=true
      - WUD_TRIGGER_DOCKER_MYUPDATER_AUTO=true
      - WUD_TRIGGER_DOCKER_MYUPDATER_PRUNE=true
[...]

PR #1064, opened 2026-05-17
fix: override getTagsPage in ECR provider to use link header for pagination

bufferpurge:
Proposed fix for handling public ECR images in WUD. Relates to issue: #1052 

Tried to build a container, but [ui/**dist/**](https://github.com/getwud/wud/blob/8242a8cb8a8e624b76a3bcc59bb3f8ca51c5424b/Dockerfile#L57) is missing from the repo. So instead I've tested by overriding `/home/node/app/dist/registries/providers/ecr/Ecr.js` in a container through a bind mount with this proposed change, which works.

### DIFF
--- a/app/registries/providers/ecr/Ecr.ts
+++ b/app/registries/providers/ecr/Ecr.ts
@@ -111,6 +111,24 @@ class Ecr extends Registry {
             return undefined;
         }
     }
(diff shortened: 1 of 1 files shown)


fmartinou:
Thank you!

PR #1076, opened 2026-06-08
Add multi-stage UI build to Dockerfile (closes need for out-of-context UI build)

terafin:
First — thank you for maintaining wud. The dual-target image management workflow is great and we use it in production daily.

## Summary

Currently `Dockerfile` expects `ui/dist/` to exist when docker build runs, but the UI build (in `ui/package.json`) isn't part of the Dockerfile itself — it has to be run separately by upstream's release CI. Anyone trying to build the image locally (from a fork, for testing, or as a downstream contributor) hits:

```
ERROR: failed to compute cache key: "/ui/dist": not found
```

## What this PR does

This is a re-submission of the change in PR #892 (originally by @steilerDev — credit preserved in the commit author), which adds a `ui-dependencies` build stage that runs `npm ci && npm run build` inside Docker, then `COPY --from=ui-dependencies` brings the built UI into the release stage. Same pattern your `app-dependencies` stage already uses for the backend.

Net effect: `docker build .` works standalone from a clean checkout. No behavioral change for anyone using upstream's release CI; the upstream Docker Hub image keeps building exactly as before.

## Test plan

- [x] `docker buildx build --platform linux/amd64,linux/arm64 .` succeeds (intarweb fork validated 2026-06-08)
- [x] Resulting image boots and serves `/health` (existing healthcheck passes)
- [x] No change to runtime image contents (same node_modules, same ui/, same Docker.entrypoint.sh)

## Why re-submitting

PR #892 is open from 2026-01 but the branch has drifted significantly behind upstream (313 additions including unrelated test/CI churn). This PR carries just the single Dockerfile + .dockerignore change, rebased onto current main, attribution preserved in the commit author header.

Closes the same root cause as #892.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -20,14 +20,6 @@ jest.config.js
 nodemon.json
 **/.DS_Store
 **/Thumbs.db
-ui/src/
-ui/public/
+
 ui/tests/
 ui/coverage/
-ui/node_modules/
-ui/package*.json
-ui/.eslintrc.js
-ui/babel.config.js
-ui/jest.config.js
-ui/vue.config.js
-ui/.browserslistrc
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,23 @@
 # Common Stage
 FROM node:24-alpine AS base
+WORKDIR /home/node/app
+
+# App dependency stage
+FROM base AS app-dependencies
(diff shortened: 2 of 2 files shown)


fmartinou:
Looks great; thank you!

PR #1079, opened 2026-06-15
Fix digest fallback when Docker Config.Image is empty

### LINKED ISSUE #1078: Digest updates for mutable non-semver tags are missed when Docker image Config.Image is empty
## Description

WUD misses digest updates for mutable `latest` tags when Docker image inspect returns an empty `Config.Image`.

This is different from #780: the compose file explicitly contains `:latest`, and the `dockercompose` trigger works once WUD marks the container as updateable.

It is also different from #724: digest watching is already enabled. The problem is that the local digest value is not populated.

## Environment

WUD version: 8.2.2

Setup:
- Docker watcher
- custom registry: registry.yandexcloud.net
- image tag: latest
- digest watching enabled implicitly for non-DockerHub non-semver images
- container image has RepoDigests and Id, but Docker image inspect returns Config.Image = ""

## Observed behavior

WUD stores the repo digest, but `image.digest.value` is undefined:

```json
"digest": {
  "watch": true,
  "repo": "sha256:...",
  "value": null
}
```

Because of that, updateAvailable falls back to tag comparison. Since both local and remote tags are latest, WUD reports 0 available updates even though the remote image digest has changed.

## Cause

In app/watchers/providers/docker/Docker.ts, the legacy branch uses:
container.image.digest.value = image.Config.Image === '' ? undefined : image.Config.Image;

For modern Docker images, image.Config.Image may be empty, while image.Id is populated and matches the local image config digest.

## Workaround tested

Changing this to use image.Id as fallback fixes detection:

container.image.digest.value = image.Config?.Image || image.Id;

After this patch, WUD correctly detects digest updates and the dockercompose trigger successfully pulls and recreates the containers.

kostjaz:
Fixes #1078.

## Summary
- Fall back to Docker image `Id` when `Config.Image` is empty while resolving the local digest for legacy manifest handling.
- Add a Docker watcher regression test for mutable non-semver tags where digest watching is enabled and Docker image inspect returns an empty `Config.Image`.

## Why

For some modern Docker images, `image.Config.Image` is empty while `image.Id` still contains the local image config digest. Leaving `image.digest.value` undefined makes WUD skip digest comparison and fall back to tag comparison, so mutable tags like `latest` can be reported as having no update even when the remote digest changed.

## Validation

From `app/`:
- `npm test -- --runInBand watchers/providers/docker/Docker.test.ts`
- `npm run build`
- `npm run lint` (passes with existing warnings)

### DIFF
--- a/app/watchers/providers/docker/Docker.test.ts
+++ b/app/watchers/providers/docker/Docker.test.ts
@@ -700,6 +700,40 @@ describe('Docker Watcher', () => {
             expect(container.image.digest.value).toBe('sha256:legacy123');
         });
 
+        test('should use image id when legacy image config digest is empty', async () => {
+            await docker.register('watcher', 'docker', 'test', {});
+            const container = {
+                image: {
+                    id: 'image123',
+                    registry: { name: 'hub' },
+                    tag: { value: 'latest' },
+                    digest: { watch: true, repo: 'sha256:abc123' },
+                },
+            };
+            const mockRegistry = {
+                getTags: jest.fn().mockResolvedValue(['latest']),
+                getImageManifestDigest: jest.fn().mockResolvedValue({
+                    digest: 'sha256:def456',
+                    created: '2023-01-01',
+                    version: 1,
+                }),
+            };
+            registry.getState.mockReturnValue({
+                registry: { hub: mockRegistry },
+            });
+            const mockLogChild = { error: jest.fn() };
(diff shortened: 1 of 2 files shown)


mlet292:
Adding visibility to this, WUD is basically useless on my home server with this issue. Thanks @kostjaz

fmartinou:
Closed in favor of #1085

PR #1085, opened 2026-07-04
Fix digest comparison for single-platform manifests resolved from a manifest list

selu:
## Summary

`getImageManifestDigest` (`Registry.ts`) misclassifies a single-platform manifest (schemaVersion 2, not a list/index) as a legacy "version 1" response, because it reads the mediaType of the *config* blob instead of the manifest's own mediaType, then returns the config blob's digest as "the digest" for that manifest.

This exact situation happens whenever Docker.ts re-resolves a container's local `RepoDigest` and that RepoDigest already points directly at a leaf/platform manifest (rather than at the top-level manifest-list digest) — which is common, and depends on the Docker storage driver / containerd snapshotter in use. Instead of re-deriving a real, comparable manifest digest through the same HEAD + `docker-content-digest` confirmation path already used for multi-arch images, the code silently swaps in the config digest.

The result: `container.image.digest.value` (local) ends up being a **config digest**, while `result.digest` (remote) is a **manifest digest** — two different kinds of object that can never be equal — so wud reports a phantom "update available" even when the running image is already current.

### Fix
- `Registry.ts`: for a single-platform manifest response, use the reference it was already fetched by (tag or digest) and the manifest's own `mediaType`, so it flows through the existing HEAD-confirmation branch and returns a genuine, comparable manifest digest (version 2) instead of a config digest mislabeled as version 1.
- `Docker.ts`: for the (now much rarer, genuinely legacy schemaVersion-1) fallback path, use `image.Config.Image || image.Id`, since `Config.Image` is blank on virtually all modern images.

### Reproduction

Confirmed against a real-world report: `ghcr.io/tricked-dev/kanidm-oauth2-manager:latest` on an arm64 host, with `wud.watch.digest` effectively enabled by default (non-semver tag, non-Docker-Hub registry). wud reported a permanent "update available" even though the running container was already on the latest image.

I reproduced this locally with podman (pulling the same public image, amd64 build) and confirmed:
- **Before fix**: local digest resolves to the image's *config* digest (`sha256:37159e9f...`) vs. remote *manifest* digest (`sha256:a8b8c6ed...`) → reported as 1 available update.
- **After fix**: both resolve to the same manifest digest (`sha256:a8b8c6ed867c437afb990988e172958ca0ab0afd892f9d2512f38f4f891046fe`) → 0 available updates.

### Related PRs

- #1079 fixes a different, narrower branch (empty `Config.Image` falling back to `undefined` instead of `image.Id`) — that fix is folded into this PR too, but it doesn't touch the bug above, since it only affects genuinely legacy schemaVersion-1 registries.
- #1038 targets the same root cause (config digest vs. manifest digest confusion) but its fix is incomplete: it corrects the branching condition but still uses `responseManifests.config.digest` as the manifest *reference* for the confirmation request, which is not a valid manifest reference and 404s (`MANIFEST_UNKNOWN`) against a real registry (verified live against GHCR). Its accompanying test doesn't catch this because the mock asserts on the `Accept` header only, never on the requested URL/reference.

### Tests

- `Registry.test.ts`: updated the single-platform-manifest test to use realistic (distinct) config vs. manifest media types, and now asserts on the actual request URLs used, not just the returned digest. Added a test for resolving a manifest fetched directly by its own digest (the RepoDigest-as-leaf-digest case).
- `Docker.test.ts`: added a regression test reproducing the `ghcr.io/tricked-dev/kanidm-oauth2-manager` scenario end-to-end through `findNewVersion` with a real `Registry` instance (mocked HTTP layer), and a test for the `Config.Image` → `image.Id` fallback.

From `app/`:
- `npx jest` — 563/563 tests pass
- `npx tsc --noEmit` — clean
- `npx eslint` on changed files — no new errors/warnings

## Test plan

- [x] `npm test` (from `app/`)
- [x] `npm run build`
- [x] Manual end-to-end reproduction with podman: pulled `ghcr.io/tricked-dev/kanidm-oauth2-manager:latest`, pointed wud's docker watcher at the podman socket (`WUD_WATCHER_LOCAL_SOCKET`), confirmed 1 false "available update" before the fix and 0 after, with identical manifest digests on both sides.

### DIFF
--- a/app/registries/Registry.test.ts
+++ b/app/registries/Registry.test.ts
@@ -165,22 +165,25 @@ test('getImageManifestDigest should return digest for application/vnd.docker.dis
     });
 });
 
-test('getImageManifestDigest should return digest for application/vnd.docker.distribution.manifest.v2+json', async () => {
+test('getImageManifestDigest should return the manifest digest (not the config digest) for a single-platform application/vnd.docker.distribution.manifest.v2+json response', async () => {
     const registryMocked = new Registry();
     registryMocked.log = log;
+    const urlsCalled: string[] = [];
     registryMocked.callRegistry = (options) => {
+        urlsCalled.push(options.url);
         if (
             options.headers.Accept ===
             'application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.index.v1+json, application/vnd.docker.distribution.manifest.v2+json, application/vnd.oci.image.manifest.v1+json'
         ) {
+            // Realistic single-platform manifest: config.mediaType is a *config*
+            // media type, never the manifest's own media type.
             return {
                 schemaVersion: 2,
                 mediaType:
                     'application/vnd.docker.distribution.manifest.v2+json',
                 config: {
-                    digest: 'digest_x',
-                    mediaType:
-                        'application/vnd.docker.distribution.manifest.v2+json',
+                    digest: 'config_digest',
+                    mediaType: 'application/vnd.docker.container.image.v1+json',
                 },
             };
         }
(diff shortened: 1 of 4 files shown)


findusl:
I'm facing the same issue, it shows my bentopdf as outdated even though it is up to date.

fmartinou:
Looks great; thank you very much for your contribution.

Issue #975, opened 2026-02-27
How to auto-update selection of containers?


MielsvS:
I've been struggling to get this to work for a looong time. I must be missing something, hopefully somebody can help me out.

**I want to periodically auto-update only explicitly marked containers,** since I don't want to auto-update databases etc.
When I create new containers and I forget to add WUD-labels I DON'T want them to get updated by default. Only when I have explicitly marked them as auto-updateable.

Can that be done?

I'm not quite sure how to use WUD_TRIGGER_DOCKER_MYTRIGGER_AUTO in combination with wud.trigger.exclude/include
Since I DO want the trigger to get automatically activated.
But I do NOT want it to be active for ALL containers.

[Here](https://linuxiac.com/how-to-keep-containers-up-to-date-with-whats-up-docker-wud/#Include_amp;_Exclude_Triggers) I read:
> When neither of the two above-mentioned [“wud.trigger.include” and “wud.trigger.exclude.”] is explicitly defined to a container, all triggers apply to every container by default.

I assume that once you use a wud.trigger.include only the included triggers are active on that container. Fine.
But does that also mean that if you use wud.trigger.include:docker.mytrigger in ONE container, mytrigger isn't active for all other containers? Since that's the behavior I'm looking for...

And while I'm at it: Can you trigger an email to inform a container has been updated?

My wud compose, in case it matters:
```
version: '3.8'
services:
  wud:
    image: getwud/wud:latest
    container_name: WUD
    read_only: false  # moet op false staan om auto-update mogelijk te maken    
    restart: unless-stopped
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:rw
      - /volume2/docker/wud2:/store:rw
    ports:
      - 3555:3000
    environment:
      # Watch alle containers
      WUD_WATCHER_DOCKER_WATCHBYDEFAULT: "true"
      WUD_WATCHER_DOCKER_SOCKET: "/var/run/docker.sock"
      WUD_WATCHER_DOCKER_WATCHALL: "false" # false (default): alleen running containers. True = ook paused, exited etc)

      WUD_TRIGGER_DOCKER_MINORUPDATE_AUTO: "false"
      WUD_TRIGGER_DOCKER_MINORUPDATE_PRUNE: "true"
      WUD_TRIGGER_DOCKER_MINORUPDATE_THRESHOLD: "minor"

      # Mailconfiguratie
      WUD_TRIGGER_SMTP_GMAIL_PASS: *****
[...]

Issue #691, opened 2025-05-22
Selectively include containers for  auto-update via docker trigger


lucascott:
My ideal setup for the Docker trigger would be that I could:

1. Manually update all containers via the UI button or API
2. For selected "less critical" containers let WUD's docker trigger automatically run the update on my behalf

Functionality 1. is possible via the `WUD_TRIGGER_DOCKER_UPDATE_AUTO=false`. 

However, functionality 2. is not supported at the moment as it's only possible to enable auto update for all containers and selectively exclude those to not update.

As a proposal it would be desirable to have a docker label assignable to those containers for whom WUD can automatically update in a similar fashion to the `wud.trigger.include`.

Happy to contribute to enable this :)

thecoshman:
I'm only starting to learn about how to configure WUD, but can't you already do this? You have a 'trigger' that does an update, which I think you can also set to not be applied by default, then include that trigger where you want it.

Salvora:
> I'm only starting to learn about how to configure WUD, but can't you already do this? You have a 'trigger' that does an update, which I think you can also set to not be applied by default, then include that trigger where you want it.

How?
if you set
`WUD_TRIGGER_DOCKER_LOCAL_AUTO=false
`

and on the container,

`wud.trigger.include=docker.local:all
`
it will still not auto update

thecoshman:
Hmm... I get what you're trying to do. At the moment though, in my head, I'm still wanting 'control' over when things update/restart. I'm sticking with using WUD more as a dashboard to show when things are out of date, but then manual arrange my updates by committing the new versions to my repo

Salvora:
I get what you mean and I also use wud as a kind of dashboard to see updates and manually update them if I  think it is a good idea. There is just one thing that bothers me is I cannot fine tune what to update. Or I can but the logic is reverse. I need to enable auto updating globally then disable the auto update for every container with labels. There are some non-critical containers I don't mind updating automatically but globally enabling this is not an option.

thecoshman:
Yeah, I think that was one the hesitations I had over WUD, I didn't like the idea of auto-updating be a default on. I'd rather WUD by default does nothing, then I opt in containers, more like Traefik. A label for WUD to pay attention to the container, a label to say auto-update can work etc.

SdotG7:
[...]

PR #1088, opened 2026-07-11
Add opt-in mode for trigger association

IARI:
## Summary
- add common `WUD_TRIGGER_{trigger_type}_{trigger_name}_INCLUDEBYDEFAULT` option
- keep existing trigger association behavior unless the option is set to `false`
- apply the option to automatic trigger execution and the container trigger association API used by manual/UI triggering
- document the new shared trigger option

## Use case
My immediate use case is a WUD setup where only a few containers should ever be updateable through a Docker Compose trigger. Most containers are managed elsewhere and should not accidentally show or run the compose trigger unless they explicitly opt in with `wud.trigger.include`.

## Related issues
This could solve the opt-in trigger/update behavior requested in #691 and #975.

## Tests
- `npm run build`
- `npm test -- --runTestsByPath triggers/providers/Trigger.test.ts api/container.test.ts --runInBand`
- `npx eslint api/container.ts api/container.test.ts triggers/providers/Trigger.ts triggers/providers/Trigger.test.ts`
- `npx jest --coverage --runInBand --testPathIgnorePatterns=triggers/providers/command`

Note: the raw full Jest suite on Windows still hits the existing command trigger `/bin/sh` assumption, so the broad local run excluded `triggers/providers/command`.


fmartinou:
Looks great; thank you very much for your contribution 👍

PR #1120, opened 2026-07-26
Prevent the up-to-date status label from overflowing its card

### LINKED ISSUE #571: [Bug] Subtitle texts overflow on home screen boxes
Here I'm attaching a screenshot to clarify this issue title:

Image

### LINKED ISSUE #760: [WEBUI BUG] "all containers are up-to-date" does not fit in box
"all containers are up-to-date" does not fit in the container box. 
Browser: Microsoft Edge (Chromium) 140.0.3485.54.

I propose changing the text to: "All up to date" or simply "Up to date".

Image

### LINKED ISSUE #851: Status text overflows  Container box
Image

blueskiestailwinds:
At screen widths at or above 1280, the container card's status button uses Vuetify's default white-space: nowrap plus a fixed button height, so the longer "all containers are up-to-date" label was clipped by the card.

uncorrected

Allow that one label to wrap, let the button grow to fit, and make the home cards fill their column so all four are equal height with 100%. It actually needs 170, but 100% is more future-proof. 

corrected

Fixes #760
Fixes #851
Fixes #571

### DIFF
--- a/ui/src/views/HomeView.vue
+++ b/ui/src/views/HomeView.vue
@@ -12,6 +12,7 @@
           >
           <br />
           <v-btn
+            class="update-status"
             size="small"
             variant="plain"
             :color="containersToUpdateCount > 0 ? 'warning' : 'success'"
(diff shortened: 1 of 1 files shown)


fmartinou:
Thank you very much 🙂!

PR #1123, opened 2026-07-31
Add nomad trigger: safely update HashiCorp Nomad-managed containers

therustyrobot:
## Summary

Adds a `nomad` trigger provider for containers whose lifecycle is managed by [HashiCorp Nomad](https://www.nomadproject.io/) rather than plain Docker/Compose.

### Why not the existing `docker`/`dockercompose` triggers?

Those triggers update a container by stopping it, removing it, and creating a replacement directly through the Docker API. That's the right approach when WUD (or the operator) is the sole owner of the container's lifecycle.

It breaks down when something else is supervising the container -- Nomad included. Nomad's client continuously reconciles the containers backing its allocations against what it expects to be running. If a `docker`/`dockercompose` trigger deletes and recreates a Nomad-managed container out from under it, Nomad sees its tracked container disappear, treats it as a task failure, and reconciles independently -- fighting whatever the trigger just created. The replacement container also isn't part of any allocation Nomad knows about, so it loses Nomad's service registration, health checks, and template-rendered secrets/config.

### What this trigger does instead

It calls [Nomad's own allocation restart API](https://developer.hashicorp.com/nomad/api-docs/allocations#restart-allocation) (`POST /v1/client/allocation/:alloc_id/restart`) and lets Nomad's own Docker driver handle the pull and recreation, the same way it would for any other restart. The allocation stays under Nomad's normal supervision throughout -- no fighting, no orphaned containers.

Nomad stamps every container it creates with a `com.hashicorp.nomad.alloc_id` label (and, on at least some versions, `com.hashicorp.nomad.task_name`) -- see `drivers/docker/driver.go` in the Nomad source. This trigger reads those to target the restart, with a fallback to parsing the task name out of the container's own name (Nomad always names containers `<task_name>-<alloc_id>`) for the Nomad versions/setups where the `task_name` label isn't present -- I observed it missing entirely on Nomad v2.0.4 while `alloc_id` was still set. If neither source yields a task name, the trigger logs a warning and does not restart anything, rather than silently falling back to restarting every task in the allocation (which could mean restarting an unrelated database sidecar in the same task group).

For the restart to actually pick up a new image, the target Nomad task needs `force_pull = true` set in its job spec -- otherwise Nomad's driver reuses whatever's already cached locally, same as any other restart would.

## Testing done

- New unit tests (`Nomad.test.ts`), 11 passing, 100% line coverage on the new file
- `npm run lint`, `npm run build` (`tsc --noEmit`), and the full existing suite (63 suites / 612 tests) all pass with no regressions
- Built the image from this branch and smoke-tested it standalone (registers correctly, respects `AUTO=false` for manual-only execution via the existing "run trigger" UI/API)
- Live end-to-end test against a real 3-node Nomad v2.0.4 cluster: opted a running container into the trigger via `wud.trigger.include`, invoked it through WUD's `/api/triggers/nomad/:name` endpoint, and confirmed via `nomad alloc status` that the *same* allocation ID restarted cleanly (0 → 1 restarts) with no duplicate/orphaned containers, and the service was healthy again afterward

## Docs

Added `docs/configuration/triggers/nomad/README.md` following the existing per-trigger doc convention, and registered it in `docs/configuration/triggers/sidebar.md`.

Happy to adjust naming/config shape if it doesn't match where you'd want this to land -- this is my first contribution here, so let me know if there's anything I should do differently.

### DIFF
--- a/app/triggers/providers/nomad/Nomad.ts
+++ b/app/triggers/providers/nomad/Nomad.ts
@@ -1,0 +1,125 @@
+import axios from 'axios';
+import Trigger from '../Trigger';
+import { ComponentConfiguration } from '../../../registry/Component';
+import { Container, fullName } from '../../../model/container';
+
+/**
+ * Restart the HashiCorp Nomad allocation/task backing a watched container.
+ *
+ * Unlike the `docker`/`dockercompose` triggers, this never touches the
+ * Docker API to stop/remove/create a container directly. Containers
+ * started by Nomad are supervised by it (service registration, health
+ * checks, template-rendered secrets, restart policy) and deleting one
+ * out from under Nomad causes it to treat the disappearance as a task
+ * failure and reconcile independently, fighting whatever just replaced
+ * it. Instead, this trigger asks Nomad itself to restart the task,
+ * which re-runs the driver's image pull (when `force_pull = true` is
+ * set on the task, image pulls are not cached) and keeps the
+ * allocation under Nomad's normal supervision throughout.
+ *
+ * Nomad stamps every container it creates with a fixed set of labels
+ * (see hashicorp/nomad drivers/docker/driver.go):
+ *   - com.hashicorp.nomad.alloc_id
+ *   - com.hashicorp.nomad.task_name
+ *   - com.hashicorp.nomad.job_name
+ *   - com.hashicorp.nomad.job_id
+ *   - com.hashicorp.nomad.task_group_name
+ * This trigger reads the first two to target the restart.
+ */
+class Nomad extends Trigger {
+    /**
+     * Get the Trigger configuration schema.
+     */
+    getConfigurationSchema() {
+        return this.joi.object().keys({
+            address: this.joi
+                .string()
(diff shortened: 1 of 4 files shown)


app/triggers/providers/nomad/Nomad.ts
```diff
@@ -0,0 +1,155 @@
...
+    /**
+     * Mask the token when logging configuration.
+     */
+    maskConfiguration(
+        configuration?: ComponentConfiguration,
+    ): ComponentConfiguration {
+        const config = configuration || this.configuration;
+        return {
+            ...config,
+            token: config.token ? '_****_' : undefined,
```
Tom-V, line 66: There is a mask function in WUD available to use
```suggestion
            token: Nomad.mask(config.token)
```

fmartinou:
It's very interesting, thanks 👍 
(although I cannot test by myself because I have no Nomad cluster around ^^)

I'll merge and fix few minor changes afterwards (as the one detected by @Tom-V )

PR #1134, opened 2026-08-05
Mask http auth password and bearer

### LINKED ISSUE #1131: Wud logs the http bearer token for http notify
Hello,

Wud seems to log the http auth token as clear text in the logs:
```
2026-08-05T07:03:36.628334727Z [2026-08-05T07:03:36.627+00:00]  INFO: whats-up-docker/trigger.http.notifyany/7 on beokpdcovm01: Register with configuration {"url":"<url here>","auto":"true","simplebody":"BEOKPDCOVM01","auth":{"bearer":"<clear text here>","type":"BEARER"},"simpletitle":"BEOKPDCOVM01"}
```
It might be better to redact this just as registry passwords

Tom-V:
I added masking of password and bearer token in the config of a http trigger. 
I also fixed typescript checking and removed the types from jsdoc as it is inferred from typescript itself

Fixes 
- #1131

### DIFF
--- a/app/triggers/providers/http/Http.test.ts
+++ b/app/triggers/providers/http/Http.test.ts
@@ -36,6 +36,28 @@ describe('HTTP Trigger', () => {
         expect(() => http.validateConfiguration(config)).toThrow();
     });
 
+    test('should mask auth password and bearer token', async () => {
+        http.configuration = {
+            url: 'https://example.com/webhook',
+            method: 'POST',
+            auth: {
+                type: 'BASIC',
+                user: 'user',
+                password: 'pass',
+                bearer: 'token',
+            },
+        };
+
+        expect(http.maskConfiguration()).toEqual({
+            ...http.configuration,
+            auth: {
+                ...http.configuration.auth,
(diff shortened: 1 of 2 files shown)


mrosnerr (approved):
Independently reproduced the 8.3.1 startup-log exposure with a synthetic file-backed HTTP bearer value. This change masks the affected Basic password and bearer fields before component registration logs them. I checked out commit 464be1bf and verified the focused HTTP tests, TypeScript compilation, and ESLint (no errors; one pre-existing @ts-nocheck warning in the test). Approved. Masking credential-bearing URL/proxy values would be worthwhile as a non-blocking follow-up.

mrosnerr:
@fmartinou This fixes the clear-text HTTP bearer/password registration logging reported in #1131. I independently reproduced the issue on 8.3.1 and validated this PR locally (focused Jest, TypeScript, and ESLint); the branch is clean and mergeable. Would appreciate maintainer review when available.

fmartinou:
LGTM 👍

Issue #615, opened 2025-02-26
[Feature request] HA autodiscovery - change device name


dimidur:
I am writing to bring to your attention an issue I encountered while using the MQTT trigger with Home Assistant autodiscovery. Currently, there is no provision to set the device ID and name for scenarios where multiple WUD instances are in use. This results in all instances appearing as a single device within Home Assistant. The device ID and name are, in fact, hardcoded as shown in this [link](https://github.com/getwud/wud/blob/25bea0cf1220a343b42b4c6a9cdef761c7cd507c/app/triggers/providers/mqtt/Hass.js#L11).

I believe it would be highly beneficial (for me 100% :D) to introduce optional configuration variables that allow users to set device IDs and names.

thevoltagesource:
[Edit again] For real this time.  Now that I tested more combinations, the device [identifier] in the JSON is important to separate the WUD instances in HA, and setting the device name changes the friendly name of the device.  The changes have been reflected below.
---

Bumping this request up to the top of the stack.  I have three Docker hosts, and I now have WUD on two of them.  I would like to see those as separate devices in Home Assistant so that update actions can be directed to the correct WUD API.  Ideally making this configurable with an environment variable like `WUD_TRIGGER_MQTT_{trigger_name}_HASS_DEVICE_NAME` or even setting the device name to the value set with  `WUD_TRIGGER_MQTT_{trigger_name}_CLIENTID` variable.  

I have tried everything I could to separte the instances with the environment variables available without luck. I have a unique name in  the watcher definition:  `WUD_WATCHER_HOUSEBRAINS_SOCKET=/var/run/docker.sock` and in the MQTT trigger definition 
`WUD_TRIGGER_MQTT_MOSQUITTO_CLIENTID=wudhb` and `WUD_TRIGGER_MQTT_MOSQUITTO_TOPIC=wudhb`

I see those customizations reflected in the HASS discovery MQTT message but it still sets the device name as "wud" which is what HASS keys off of to group the entities.

{
  "unique_id": "wudhb_housebrains_wud",
  "default_entity_id": "update.wudhb_housebrains_wud",
  "name": "What's Up Docker",
 "device": {
    "identifiers": [
      "wud".  <---- ***This needs to be unique per WUD instance. ***
    ], 
    "manufacturer": "wud",
    "model": "wud",
    "name": "wud",   <---- ***This could be unique per WUD instance for easy Identification***
    "sw_version": "8.2.2"
  },
[...]

PR #1135, opened 2026-08-05
Add deviceid and devicename to MQTT HASS

Tom-V:
I added device id and device name as optional configurable options, it still defaults to wud

I don't use hass or mqtt locally, but the changes are minimal, however I did rework the whole hass file implementation to reuse code instead of having everything duplicated, so it should be easier to maintain. The tests back up that nothing changed. 
As usual I also added typescript typings.

Added a feature to add
- #615


fmartinou:
Thanks 🙂

PR #1138, opened 2026-08-10
Fix login redirect issues with subpath handling in WUD

balaji-g42:
This pull request introduces improvements to how public URLs and paths are constructed and handled across both the backend and frontend, ensuring correct behavior when a base path is configured. It adds comprehensive tests for these scenarios and updates service worker registration and PWA configuration to better handle dynamic base paths.

**Backend URL Construction and Testing:**
- Enhanced the `getPublicUrl` function in `app/configuration/index.ts` to correctly append the server base path to the public URL or request origin, and updated its logic to handle cases where both `WUD_PUBLIC_URL` and `WUD_SERVER_BASEPATH` are set.
- Added new tests in `app/configuration/index.test.ts` to verify `getPublicUrl` behavior with different combinations of environment variables and request origins.

**Frontend URL Construction and Testing:**
- Updated the `url` function in `ui/src/services/base.ts` to ensure paths are joined correctly, always inserting a slash between the base path and the requested path, and preventing double slashes.
- Added tests in `ui/tests/services/base.spec.ts` to verify the correct joining of base paths and paths in different scenarios, including with and without trailing slashes.

**Service Worker and PWA Configuration:**
- Modified service worker registration in `ui/src/registerServiceWorker.ts` to use the updated `url` function for determining the service worker script path, ensuring compatibility with dynamic base paths.
- Updated `ui/vue.config.js` to exclude `index.html` from the service worker precache, as it is generated dynamically per request with base path injection.

### DIFF
--- a/app/configuration/index.test.ts
+++ b/app/configuration/index.test.ts
@@ -96,6 +96,44 @@ test('getServerConfiguration should return configured api (new vars)', async () 
     });
 });
 
+test('getPublicUrl should return request origin when no basepath is configured', async () => {
+    delete configuration.wudEnvVars.WUD_PUBLIC_URL;
+    delete configuration.wudEnvVars.WUD_SERVER_BASEPATH;
+    const req = {
+        protocol: 'https',
+        headers: { host: 'wud.example.com' },
+    };
+    expect(configuration.getPublicUrl(req)).toStrictEqual(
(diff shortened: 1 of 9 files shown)


fmartinou:
Looks great; thank you very much 👍

PR #1141, opened 2026-08-18
fix(http): mask sensitive trigger configuration

mrosnerr:
## Summary

- mask HTTP trigger URLs and proxy values before component registration logs them
- mask nested Basic and Bearer authentication fields
- add regression coverage for configurations with and without optional authentication

## Why

`Component.register()` logs each provider's configuration at INFO after calling `maskConfiguration()`. The base implementation returns the configuration unchanged, and the HTTP trigger did not override it.

Because `__FILE` values are resolved before component registration, an HTTP trigger configured with a file-backed password or bearer token could therefore write the resolved credential to the startup log. URLs and proxy values may also contain embedded credentials.

This change follows the masking pattern already used by other trigger and registry providers.

## Verification

- `npx jest triggers/providers/http/Http.test.ts --runInBand`
- `npx eslint triggers/providers/http/Http.ts triggers/providers/http/Http.test.ts` (existing `@ts-nocheck` warnings only)
- `npx tsc --noEmit`
- full app test suite: 62 suites / 605 tests passed

### DIFF
--- a/app/triggers/providers/http/Http.test.ts
+++ b/app/triggers/providers/http/Http.test.ts
@@ -36,6 +36,50 @@ describe('HTTP Trigger', () => {
         expect(() => http.validateConfiguration(config)).toThrow();
     });
 
+    test('should mask sensitive configuration data', async () => {
+        http.configuration = {
+            url: 'https://example.com/hooks/http-url-canary',
+            auth: {
+                type: 'BEARER',
+                user: 'http-user-canary',
+                password: 'http-password-canary',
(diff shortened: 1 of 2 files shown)


Tom-V:
This seems like a duplicate of #1134

mrosnerr:
Closing in favor of #1134, which predates this PR and fixes the same password/bearer registration-log exposure. I independently reproduced and validated #1134 and submitted an approving review there. The additional URL/proxy masking from this PR can be handled as follow-up hardening.

PR #1142, opened 2026-08-20
Preserve update results across registry errors

### LINKED ISSUE #1137: Registry errors are reported as "no update available": watchContainer deletes the previous result and updateAvailable falls back to false
## Summary

When a registry query fails for a container (in my case HTTP 429 from ghcr.io), WUD does not mark that container as "unknown" — it reports it as **up to date**, and it **deletes the previously-known-good result from the store**.

`watchContainer()` deletes `container.result` *before* the registry call and never restores it on the error path. The `updateAvailable` getter returns a hard `false` whenever `result` is `undefined`. The wiped container is then persisted with `storeContainer.updateContainer()`, which removes and re-inserts the record wholesale.

The net effect is that the "available updates" total silently becomes a floor rather than a total, for the whole window between a failed run and the next successful one (an hour on the default `0 * * * *` cron). The error count is printed in the cron log line, but nothing downstream distinguishes "checked, up to date" from "check failed".

This is separate from #1136, which covers *why* the registry calls fail (unbounded concurrency / request volume). This issue is about what WUD does with the failure. It applies to any transient registry error, not only 429s.

## Environment

- WUD **8.3.1**, image `getwud/wud:8.3.1`
- One Docker watcher, local socket, default cron `0 * * * *`, `watchevents` default (true)
- **71 containers** watched
- Registries in play: `ghcr.io` (linuxserver/\*, immich-app/\*) and Docker Hub
- `WUD_REGISTRY_GHCR_PUBLIC_USERNAME` / `_TOKEN` configured (a GitHub classic PAT with `read:packages`), confirmed loaded in the startup log

## Steps to reproduce

Any transient registry failure reproduces it; a 429 is just the easy way to get one.

1. Watch a container that has an update available and let one scan succeed. Confirm `updateAvailable: true` in `GET /api/containers` and in the UI.
2. Make that container's registry fail for one scan — e.g. block egress to the registry host from the WUD container, or point it at an unreachable host.
3. Trigger a watch (`POST /api/watchers/{name}/watch`, or wait for the cron).
4. Observe the log: `Error when processing (...)` at WARN, and `Cron finished (N containers watched, 1 errors, 0 available updates)`.
5. `GET /api/containers` now returns `"updateAvailable": false` for that container, and the stored `result` is gone. The UI counts it as up to date.
6. Restore connectivity; the next successful run restores the correct state.

## Evidence

Same watcher, same 71 containers, same day. The only variable is how many containers errored:

| Time | Run type | errors | available updates |
|---|---|---|---|
| 16:01:17 | cron | 8 | 5 |
| 16:04:07 | event-triggered | 0 | **11** |
| 16:04:31 | event-triggered | 0 | **10** |
| 16:04:43 | event-triggered | 0 | **10** |
| 16:05:33 | event-triggered | 0 | **10** |
| 17:01:13 | cron | 10 | 5 |

All 8 errors in the 16:01 run were `429 - Too Many Requests` from ghcr.io. In the 17:01 run, 8 of the 10 were the same 429s; the other two were a 404 (`recyclarr`) and a 504 (`sound-redis`).

Every failing image in these runs was on ghcr.io. Docker Hub images in the same runs never errored — consistent with Docker Hub being the one registry that overrides `shouldWatchDigest()` to default `false` ([`Hub.ts#L33`](https://github.com/getwud/wud/blob/f249e06d95176de01e16faeb90072efe3db4de7d/app/registries/providers/hub/Hub.ts#L33)) while the base returns `true` ([`Registry.ts#L53`](https://github.com/getwud/wud/blob/f249e06d95176de01e16faeb90072efe3db4de7d/app/registries/Registry.ts#L53)).

I am not claiming to know why the top-of-hour runs are the ones that get throttled — the event-triggered runs go through the same `watch()` code path and issue the same requests, so I can't attribute the difference to request volume alone. That question belongs to #1136. What this table shows is the relationship this issue is about: **the reported update total moves inversely with the error count**, from 10–11 down to 5.

## What I expected vs what happened

**Expected:** a container whose registry check failed is reported as *unknown / not checked*. Its last known state is preserved until a check succeeds, and it is not counted as up to date.

**Actual:** it is reported as `updateAvailable: false` — indistinguishable from a successful "up to date" check to every consumer of the API, UI, MQTT/HASS sensors, Prometheus labels and triggers — and its previous `result` is deleted from the store.

## Suspected cause

At tag `8.3.1` (commit `f249e06`):

1. [`app/watchers/providers/docker/Docker.ts#L547-L570`](https://github.com/getwud/wud/blob/f249e06d95176de01e16faeb90072efe3db4de7d/app/watchers/providers/docker/Docker.ts#L547-L570) — `result` is deleted before the try block; the catch sets only `error`, never restores `result`:

```ts
// Reset previous results if so
delete containerWithResult.result;
delete containerWithResult.error;
...
} catch (e: any) {
    logContainer.warn(`Error when processing (${e.message})`);
    containerWithResult.error = { message: e.message };
}
```

2. [`app/model/container.ts#L184-L190`](https://github.com/getwud/wud/blob/f249e06d95176de01e16faeb90072efe3db4de7d/app/model/container.ts#L184-L190) — the computed getter has no "unknown" state:

```ts
get(this: Container) {
    if (this.image === undefined || this.result === undefined) {
        return false;
    }
```

3. [`Docker.ts#L890`](https://github.com/getwud/wud/blob/f249e06d95176de01e16faeb90072efe3db4de7d/app/watchers/providers/docker/Docker.ts#L890) → [`app/store/container.ts#L47-L64`](https://github.com/getwud/wud/blob/f249e06d95176de01e16faeb90072efe3db4de7d/app/store/container.ts#L47-L64) — `updateContainer()` does `.remove()` then `.insert()`, so the wiped `result` is what gets persisted.

4. [`Docker.ts#L490-L503`](https://github.com/getwud/wud/blob/f249e06d95176de01e16faeb90072efe3db4de7d/app/watchers/providers/docker/Docker.ts#L490-L503) — the summary derives `errors` and `available updates` from the same array with two independent filters, so an errored container lands in the first and is excluded from the second.

To be fair to the current design: this is transient, not permanent. `addImageDetailsToContainer` bypasses the store cache when `error !== undefined` (`Docker.ts#L756-L763`), so the next run re-queries and self-heals. The problem is the window in between, and that nothing outside the log line indicates the number is unreliable.

## Why it matters

The failure is quiet rather than loud. The cron line does print the error count, but from the code every other surface reads the same boolean and carries no error signal:

- the UI home view counts `containers.filter(c => c.updateAvailable)` and renders "all containers are up-to-date" at zero;
- `Hass.ts` derives the `update_count` sensor and the update binary sensor from `getContainers({updateAvailable: true})`;
- Prometheus emits the container with `update_available="false"`; there is no watch-error metric;
- triggers gate solely on `container.updateAvailable`, so a failed check fires no notification of any kind.

(I have directly observed only the log lines and the API/UI state on my deployment; the MQTT and Prometheus consequences follow from the same getter but I have not measured them.)

There may also be a recovery-side effect I have not confirmed: because the error wipes the stored `result`, the next successful run compares `undefined` against the new tag, so `resultChanged()` is true and a trigger can re-fire for an update the user was already notified about, even with the default `once: true`. If that holds it may be relevant to #1037.

## Possible fixes

Anything that stops "check failed" collapsing into "no update" would do. Options, roughly in order of invasiveness:

1. Don't delete `result` up front — write the new result on success, and leave the previous one intact on error (surfacing it as stale).
2. Exclude errored containers from the "up to date" population rather than counting them as `updateAvailable: false` — in the cron summary, the UI count, and the HASS sensor.
3. Longer term, make the state tri-valued (`update available` / `up to date` / `check failed`) instead of a boolean.

## Related

- #1136 — root cause of the failures on my deployment (unbounded registry concurrency, GHCR 429s). This issue is about the handling of those failures, not their cause.
- #923 — describes the same `result`-deletion mechanism; its specific defect (a `TypeError` in `resultChanged` that permanently muted triggers) appears fixed in 8.3.1 by optional chaining, and its proposed fix would not address the update count.
- #989 — another user reporting that these errors "prevent the latest image versions from appearing in the app". Same symptom, different trigger.
- #266 — Home Assistant update count resetting to 0; mechanically consistent with this.
- #419, #572 — registry errors are not surfaced anywhere the operator will see them.

---

magrhino:
Registry failures currently remove a container’s previous update result before persisting the error, making a known available update appear up to date until a later successful scan. This preserves the last-known-good result while continuing to expose the registry error, without changing the API or UI model.

### Changes

- Keep the previous result when a Docker registry request fails
- Rebuild errored cached containers from current Docker and registry metadata while retaining that result across consecutive failures
- Prevent simple and batch triggers from acting on stale errored reports
- Cover repeated failures, metadata recovery, successful recovery, and trigger suppression

### Testing

- `cd app && npm test -- watchers/providers/docker/Docker.test.ts triggers/providers/Trigger.test.ts --runInBand --coverage=false --silent` — 98 tests passed
- `cd app && npm test -- --runInBand --silent --coverage=false` — 605 tests passed
- `cd app && npm run lint` — 0 errors (130 existing warnings)
- `cd app && npm run build`

Fixes #1137

### DIFF
--- a/app/triggers/providers/Trigger.test.ts
+++ b/app/triggers/providers/Trigger.test.ts
@@ -105,6 +105,15 @@ const handleContainerReportTestCases = [
         updateAvailable: false,
         semverDiff: 'major',
     },
+    {
+        shouldTrigger: false,
+        threshold: 'all',
+        once: false,
+        changed: false,
+        updateAvailable: true,
+        semverDiff: 'major',
+        error: { message: 'Registry error' },
+    },
 ];
 
 test.each(handleContainerReportTestCases)(
@@ -127,6 +136,7 @@ test.each(handleContainerReportTestCases)(
                     kind: 'tag',
                     semverDiff: item.semverDiff,
                 },
+                ...(item.error && { error: item.error }),
             },
         });
         if (item.shouldTrigger) {
@@ -205,6 +215,15 @@ const handleContainerReportsTestCases = [
         updateAvailable: false,
         semverDiff: 'major',
     },
+    {
+        shouldTrigger: false,
+        threshold: 'all',
+        once: false,
+        changed: false,
+        updateAvailable: true,
+        semverDiff: 'major',
+        error: { message: 'Registry error' },
+    },
 ];
 
 test.each(handleContainerReportsTestCases)(
@@ -228,6 +247,7 @@ test.each(handleContainerReportsTestCases)(
                         kind: 'tag',
                         semverDiff: item.semverDiff,
                     },
+                    ...(item.error && { error: item.error }),
                 },
             },
         ]);
--- a/app/triggers/providers/Trigger.ts
+++ b/app/triggers/providers/Trigger.ts
@@ -162,6 +162,7 @@ class Trigger extends Component {
         // Filter on changed containers with update available and passing trigger threshold
         if (
             (containerReport.changed || !this.configuration.once) &&
+            containerReport.container.error === undefined &&
             containerReport.container.updateAvailable
         ) {
             const logContainer =
@@ -221,6 +222,7 @@ class Trigger extends Component {
                 )
                 .filter(
                     (containerReport) =>
+                        containerReport.container.error === undefined &&
                         containerReport.container.updateAvailable,
                 )
                 .filter((containerReport) =>
--- a/app/watchers/providers/docker/Docker.test.ts
+++ b/app/watchers/providers/docker/Docker.test.ts
@@ -482,24 +482,38 @@ describe('Docker Watcher', () => {
             expect(event.emitContainerReport).toHaveBeenCalled();
         });
 
-        test('should handle container processing error', async () => {
-            const container = { id: 'test123', name: 'test' };
+        test('should preserve previous result across errors and replace it on recovery', async () => {
+            const previousResult = { tag: '2.0.0' };
+            const container = {
+                id: 'test123',
+                name: 'test',
+                result: previousResult,
+            };
             const mockLogChild = { warn: jest.fn(), debug: jest.fn() };
             const mockLog = { child: jest.fn().mockReturnValue(mockLogChild) };
             docker.log = mockLog;
             docker.findNewVersion = jest
                 .fn()
-                .mockRejectedValue(new Error('Registry error'));
+                .mockRejectedValueOnce(new Error('Registry error'))
+                .mockRejectedValueOnce(new Error('Registry error'))
+                .mockResolvedValueOnce({ tag: '3.0.0' });
             docker.mapContainerToContainerReport = jest
                 .fn()
                 .mockReturnValue({ container, changed: false });
 
             await docker.watchContainer(container);
-
-            expect(mockLogChild.warn).toHaveBeenCalledWith(
-                expect.stringContaining('Registry error'),
-            );
+            expect(container.result).toBe(previousResult);
             expect(container.error).toEqual({ message: 'Registry error' });
+
+            await docker.watchContainer(container);
+            expect(container.result).toBe(previousResult);
+            expect(container.error).toEqual({ message: 'Registry error' });
+
+            await docker.watchContainer(container);
+
+            expect(mockLogChild.warn).toHaveBeenCalledTimes(2);
+            expect(container.result).toEqual({ tag: '3.0.0' });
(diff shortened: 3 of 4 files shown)


fmartinou:
LGTP 👌

Issue #1136, opened 2026-08-05
GHCR 429s for LinuxServer images: bound registry concurrency and avoid unnecessary tag enumeration


magrhino:
## Problem

WUD can generate a burst of registry requests large enough for GHCR to return HTTP 429 during a scheduled scan. Reports of this behavior commonly affect groups of `ghcr.io/linuxserver/*` images at the same time, while other checks may still succeed.

The current WUD 8.3.1 scan path appears to amplify registry request volume:

1. The Docker watcher checks containers concurrently with `Promise.all(...)`.
2. `findNewVersion()` calls `registryProvider.getTags(container.image)` even for a non-semver tag such as `latest`, where the useful operation is a digest comparison for the current tag.
3. `getTags()` follows every registry pagination link.
4. Digest watching then adds manifest requests.
5. Registry requests surface HTTP 429 immediately; there is no bounded retry/backoff that honors `Retry-After`.

The result can be a large concurrent burst of tag-list pages and manifest calls, leaving throttled containers unchecked until a later successful scan.

## Why LinuxServer images are an important example

LinuxServer.io is one of the most widely used Docker image publishers, so this workload is worth accounting for explicitly rather than treating it as an unusual edge case.

Many LinuxServer repositories have deep tag histories. For example, the public GHCR endpoint for `linuxserver/bazarr` returns a pagination link even when requesting 1,000 tags, meaning WUD must make additional requests to enumerate the complete tag list. Watching several LinuxServer images multiplies that pagination cost.

LinuxServer images expose the problem disproportionately, but the underlying behavior can affect any popular publisher or deployment watching many images from one registry.

## Expected behavior

A scheduled scan should avoid overwhelming one registry, and a transient 429 should not leave a group of popular images unchecked without a bounded retry.

## Suggested fix / configurable settings

Could WUD implement the following?

- Skip full tag enumeration when the configured tag is non-semver and only the current-tag digest is needed.
- Add a configurable per-registry concurrency limit, with a conservative default.
- Retry HTTP 429 with bounded exponential backoff and jitter, honoring `Retry-After` when present.
- Preserve successful container results while retrying or deferring only throttled checks.

A targeted per-container check or fair retry ordering could complement this, but reducing the request burst is the primary fix.

## Reproduction outline

1. Configure WUD 8.3.1 to watch many GHCR images, including several `ghcr.io/linuxserver/*:latest` images with digest watching.
2. Run a scheduled watcher scan.
[...]

PR #1143, opened 2026-08-20
fix(docker): skip tag listing for non-semver images

magrhino:
WUD currently enumerates every registry tag for all watched images, including non-semver tags such as `latest`. `getTagCandidates()` discards every non-semver candidate afterward, so repositories with long tag histories can generate multiple avoidable paginated GHCR requests before the current-tag digest check.

This change only fetches tags for semver update checks. Non-semver images keep the existing current-tag digest path, while semver images continue to enumerate candidates. The scope is intentionally limited to removing the request amplification demonstrated in the issue.

## Change summary

- Skip `getTags()` for non-semver images.
- Preserve digest resolution for current mutable tags.
- Add regression coverage proving `latest` skips enumeration while semver checks still enumerate tags.

## Testing

- `npm --prefix app test -- --runInBand --coverage=false watchers/providers/docker/Docker.test.ts` — 56 passed.
- `npm --prefix app test -- --runInBand --silent` — 603 passed.
- `npm --prefix app run lint` — 0 errors.
- `npm --prefix app run build` — passed.

Addresses #1136

### DIFF
--- a/app/watchers/providers/docker/Docker.test.ts
+++ b/app/watchers/providers/docker/Docker.test.ts
@@ -596,7 +596,7 @@ describe('Docker Watcher', () => {
             const container = {
                 image: {
                     registry: { name: 'hub' },
-                    tag: { value: '1.0.0' },
+                    tag: { value: '1.0.0', semver: true },
                     digest: { watch: false },
                 },
             };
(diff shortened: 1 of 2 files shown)


fmartinou:
Hi,

It makes sense 👍 .

Issue #624, opened 2025-03-07
[Question] API requests with OIDC auth


noenik:
When using OIDC for authentication I'm having issues authenticating API requests.

I am using Keycloak and have configured a client for WUD and can successfully authenticate in the UI with no issues.

However, when I try using the API I cannot get it to work.

Ideally, I'd like to authenticate as a service account since this is going to be used by an automated service. So I've created an service account in Keycloak and gotted an access token for it:

```
curl -H 'Content-Type: application/x-www-form-urlencoded' \
        'https://our.keycloak.domain/realms/ourrealm/protocol/openid-connect/token' \
        -d 'grant_type=client_credentials' \
        -d 'client_id=myserviceaccount' \
        -d 'client_secret=serviceaccountsecret'
```

and then I try using that token to call the API, e.g.

`curl -H 'Authorization: Bearer $TOKEN' https://wud.domain/api/containers`

I will always get a 401 response. And in the WUD logs I see simply
```
wud  | 08:36:16.944  WARN whats-up-docker/authentication.oidc.keycloak: Error when validating the user access token (invalid_token (Token verification failed))
wud  | 08:36:16.944  WARN whats-up-docker/authentication.oidc.keycloak: Bearer token is invalid
```
(which is the same I get when sending no token at all)

I've tried getting a token for my personal user using the wud client in Keycloak with the same result.

As a temporary workaround I've created basic auth credentials for this service and that works fine.

I feel the documentation is bit lacking in this area. Is there something I am missing? 

Thanks in advance

Toxicterror:
did no one get this working?

PR #1146, opened 2026-08-23
fix(oidc): pass skipSubjectCheck for bearer token userinfo validation

### LINKED ISSUE #1145: OIDC bearer-token API authentication always fails with "expectedSubject" must be a string
### Summary

With an OIDC authentication provider configured, the UI login flow works, but authenticating an API request with `Authorization: Bearer <access_token>` always returns 401 — even when the token is valid and accepted by the identity provider's own userinfo endpoint.

### Version

- WUD (current `main`, openid-client `6.8.2`)
- IdP: Authentik 2026.8.0

### Reproduce

1. Configure an OIDC provider (`WUD_AUTH_OIDC_<NAME>_DISCOVERY` / `_CLIENTID` / `_CLIENTSECRET`).
2. Obtain a valid access token for that client (e.g. via the authorization code flow, with `openid email profile` scope).
3. Confirm the token is good against the IdP:
   `curl -H "Authorization: Bearer $T" https://<idp>/application/o/userinfo/` → **200**
4. Send the same token to WUD:
   `curl -H "Authorization: Bearer $T" https://<wud>/api/containers` → **401**

WUD logs, in the same second:
```
Error when validating the user access token ("expectedSubject" must be a string)
Bearer token is invalid
```

Step 3 is the control: the token is valid at the IdP, so the 401 is not the token, the client configuration, or the network.

### Cause

In `app/authentications/providers/oidc/Oidc.ts`:

- `OidcStrategy.authenticate()` extracts the bearer token and calls `verify(accessToken, done)`.
- `verify()` calls `getUserFromAccessToken(accessToken)` with **no** `claim` argument (the authorization-code path passes `tokenSet.claims()`; the bearer path has nothing to pass).
- `getUserFromAccessToken()` then calls `fetchUserInfo(config, accessToken, claim?.sub)`, so `expectedSubject` is `undefined`.
- `oauth4webapi`'s `processUserInfoResponse()` requires `expectedSubject` to be either a string or its `skipSubjectCheck` sentinel, and throws `"expectedSubject" must be a string` for anything else, including `undefined`.

This is a regression from the `openid-client` v4 → v6 migration: v4's `client.userinfo(accessToken)` had no such parameter at all.

I verified this end-to-end against a real Authentik instance: identical access token, identical WUD build minus one line — before: `401` + the error above; after (see linked PR): `200`.

Relates to #624 (same "API auth doesn't work" symptom, different/older error — likely pre-6.x `openid-client`).

HugoKNL:
Fixes #1145
Relates to #624

## Problem

OIDC bearer-token API authentication (`Authorization: Bearer <token>`) always fails with 401, logging:
```
Error when validating the user access token ("expectedSubject" must be a string)
```
even for a token the IdP itself accepts. See #1145 for full repro.

## Cause

`getUserFromAccessToken()` calls `client.fetchUserInfo(config, accessToken, claim?.sub)`. On the bearer-token path there's no `claim` (no authorization-code exchange happened for this request), so `expectedSubject` is `undefined` — which `oauth4webapi` rejects outright rather than treating as "skip the check". The authorization-code/browser-login path is unaffected since it always supplies a real `sub` from the ID token.

## Fix

Pass the library's own sentinel for "nothing to check the subject against":

```diff
-    claim?.sub,
+    claim?.sub ?? client.skipSubjectCheck,
```

This keeps the subject cross-check on the authorization-code path (where a trusted `sub` is available) and explicitly skips it only on the bearer path (where there's nothing to compare against — the token's validity is already established by the userinfo call succeeding at all).

## Testing

- Added unit tests asserting `fetchUserInfo` is called with `client.skipSubjectCheck` when no claim is passed, and with the real `sub` when one is (previously untested — the existing suite fully mocks `openid-client`, so it couldn't catch the missing argument).
- `npm test` — all tests pass.
- Verified live against a real Authentik 2026.8.0 instance: same access token, same WUD build, only this one-line change — before: `401` / `"expectedSubject" must be a string`; after: `200`.

## Optional follow-up (out of scope here)

WUD validates bearer tokens via a live call to the IdP's `/userinfo` endpoint rather than local JWT/JWKS verification. That's spec-valid but has real limitations worth a separate discussion: no audience (`aud`) check, a per-request runtime dependency on the IdP, and poor fit for client-credentials/service-account tokens (confirmed while testing this fix — Authentik returns `403 insufficient_scope` for `/userinfo` with a client-credentials token, since it's not tied to an end-user). This PR only fixes the crash in the existing mechanism.

### DIFF
--- a/app/authentications/providers/oidc/Oidc.test.ts
+++ b/app/authentications/providers/oidc/Oidc.test.ts
@@ -177,3 +177,35 @@ test('getUserFromAccessToken should return unknown for missing email', async () 
     const user = await oidc.getUserFromAccessToken('token');
     expect(user).toEqual({ username: 'unknown' });
 });
+
+test('getUserFromAccessToken should skip the subject check when called without a claim (bearer token path)', async () => {
+    oidc.configuration = { ...configurationValid, ttl: -1 };
+    (oidc as any).cachedConfig = mockConfig;
+    (client.fetchUserInfo as jest.Mock).mockResolvedValue({
+        email: 'bearer@example.com',
+    });
+
+    await oidc.getUserFromAccessToken('token');
+
+    expect(client.fetchUserInfo).toHaveBeenCalledWith(
+        mockConfig,
+        'token',
+        client.skipSubjectCheck,
+    );
+});
+
+test('getUserFromAccessToken should check the subject when called with a claim (authorization code path)', async () => {
+    oidc.configuration = { ...configurationValid, ttl: -1 };
+    (oidc as any).cachedConfig = mockConfig;
+    (client.fetchUserInfo as jest.Mock).mockResolvedValue({
+        email: 'code@example.com',
+    });
+
+    await oidc.getUserFromAccessToken('token', { sub: 'abc' } as any);
+
+    expect(client.fetchUserInfo).toHaveBeenCalledWith(
+        mockConfig,
+        'token',
+        'abc',
+    );
+});
--- a/app/authentications/providers/oidc/Oidc.ts
+++ b/app/authentications/providers/oidc/Oidc.ts
@@ -271,7 +271,7 @@ class Oidc extends Authentication {
         const userInfo = await client.fetchUserInfo(
             config,
             accessToken,
-            claim?.sub,
+            claim?.sub ?? client.skipSubjectCheck,
         );
 
         // check the usernameclaim, if it doesn't exist, try to use the email and log a warning


HugoKNL:
The Travis CI failure here looks unrelated to this change — worth a restart rather than a fix.

All 3 failing scenarios are in the `gitlab_test` e2e case, which queries the live `registry.gitlab.com` for `gitlab-org/gitlab-runner` tags and expects to see `v16.1.0`. The failure is:
```
And response body path $.result.tag should be v16.1.0
  actual: "v16.0.0", error: "Request failed with status code 401"
```
i.e. the registry call itself 401'd, so WUD never got the tag list and just reported back the container's current tag instead of the expected update. This diff only touches `app/authentications/providers/oidc/`, nothing in the registry/e2e code path.

A couple of data points suggesting this is a transient CI-side issue rather than a standing break:
- The base commit this branch is built on (`188b8e4`, "Rewrite documentation") passed this same e2e suite against the same live registry about 2 hours before this build ran.
- Right now, GitLab's registry and the standard anonymous-pull auth flow both work fine:
  ```
  curl https://gitlab.com/jwt/auth?service=container_registry&scope=repository:gitlab-org/gitlab-runner:pull
  → 200, valid token
  curl -H "Authorization: Bearer <token>" https://registry.gitlab.com/v2/gitlab-org/gitlab-runner/tags/list
  → 200, full tag list
  ```

I don't have permission to restart the Travis build myself — could a maintainer kick off a rerun? Happy to dig further if it fails again on retry.

fmartinou:
LGTM 👍 
Thanks for your contribution!

PR #1153, opened 2026-08-26
fix(registry): limit requests and retry rate limits

### LINKED ISSUE #1136: GHCR 429s for LinuxServer images: bound registry concurrency and avoid unnecessary tag enumeration

magrhino:
WUD starts container checks concurrently, so semver images assigned to the same registry can issue a concentrated burst of tag-list and manifest requests. A transient HTTP 429 currently fails immediately, leaving those images unchecked until a later scan.

Production diagnosis for #1136 observed 51 watches start within 18 ms, followed by 10 GHCR tag-list requests within roughly 3 ms; nine returned HTTP 429 within the next second. The same burst pattern repeated on the configured scan schedule. #1143 removed unnecessary tag enumeration for non-semver images, but semver checks still need bounded registry traffic and retry handling.

## Change summary

- Limit each configured registry instance to two active requests by default, with FIFO queuing and an optional `WUD_REGISTRY_{REGISTRY_TYPE}_{REGISTRY_NAME}_CONCURRENCY` override.
- Retry only HTTP 429 responses, with two bounded retries that honor `Retry-After` values up to 60 seconds and otherwise use jittered exponential backoff.
- Release request permits before retry delays, preserve non-429 behavior, and keep limits isolated between registry instances.
- Preserve authenticated and anonymous provider validation, including Custom registry URL requirements and Quay authentication-endpoint rate limits.
- Document the new registry setting and retry behavior in the registry and GHCR configuration guides and changelog.

## Testing

- `npm --prefix app test -- --runInBand --coverage=false registries` — 25 suites / 287 tests passed.
- `npm --prefix app test -- --runInBand --silent --coverage=false --forceExit` — 99 suites / 1,003 tests passed.
- `npm --prefix app run build` — passed.
- `npm --prefix app run lint` — 0 errors (161 existing warnings).
- `npm --prefix website run build` — passed.
- Prettier check for affected backend files and the changelog — passed; migrated registry docs retain the existing Docusaurus formatting.
- `git diff --check` — passed.

Closes #1136

### DIFF
--- a/app/registries/Registry.test.ts
+++ b/app/registries/Registry.test.ts
@@ -1,4 +1,5 @@
 // @ts-nocheck
+import axios from 'axios';
 import log from '../log';
 
 jest.mock('axios');
@@ -339,7 +340,6 @@ test('getImageManifestDigest should throw when no digest found', async () => {
 });
 
 test('callRegistry should call authenticate', async () => {
-    const { default: axios } = await import('axios');
     axios.mockResolvedValue({ data: {} });
     const registryMocked = new Registry();
     registryMocked.log = log;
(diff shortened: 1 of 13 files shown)


fmartinou:
Hi,

Thanks for the PR. I just landed a significant refactor for Registry support and docs.
Could you please rebase your changes on the latest main branch?

magrhino:
@fmartinou apologies for delay, I have rebased docs to the new Docusaurus structure and resolved registry V2 test conflicts.

fmartinou:
Thanks! 👍

Issue #938, opened 2026-02-16
wud.tag.transform broken by 8.2.0


jfdahms:
I have the following compose.yml.
```
---
services:
  sabnzbd:
    image: lscr.io/linuxserver/sabnzbd:4.5.5-ls239
    labels:
      - 'wud.tag.include=^\d+\.\d+\.\d+-ls\d+$$'
      - 'wud.tag.transform=^(\d+\.\d+\.\d+)-ls\d+$$ => $$1'
```
v8.1.1 said no upgrade is available.
v8.2.0 says an upgrade to 4.5.5-ls244 is available.
My tags are specifically designed to ignore the -lsnnn portion of the version.

jfdahms:
<img width="948" height="503" alt="Image" src="https://github.com/user-attachments/assets/552ab0d6-d924-43e4-93e2-a54b5c9bc5bd" />

meuter:
I also have several regular expressions that are ignored

p-monteiro:
I think there is a general misunderstaing on how include and exclude work, maybe on my part. But from what I gathered in the source code, this is how it works:

wud.tag.include - Used to filter which images in the registry are possible candidates
wud.tag.transform - Used to filter which part of the tag is used for version comparison

Given your configuration:
```yaml
services:
  sabnzbd:
    image: lscr.io/linuxserver/sabnzbd:4.5.5-ls239
    labels:
      - 'wud.tag.include=^\d+\.\d+\.\d+-ls\d+$$'
      - 'wud.tag.transform=^(\d+\.\d+\.\d+)-ls\d+$$ => $$1'
```

All X.Y.Z-lsW tags (include) are candidates for possible upgrading (4.5.5-ls239, 4.5.5-ls244, etc.)
The X.Y.Z (transform) is used for comparison, which includes greater OR equal semvers.
So, 4.5.5-ls239 and 4.5.5-ls244 are valid upgrade paths (4.5.5 >= 4.5.5). 
The candidates are then sorted based on when they were released.
As 244 is more recent than 239, WUD will suggest it as a valid update.

jfdahms:
@p-monteiro I don't understand why we would use >= instead of > when checking for an upgrade.  Transform tells us how to change the version before comparison. In my example they both transform to 4.5.5. Therefore they are the SAME version and no upgrade is required.

[...]

PR #1163, opened 2026-08-29
fix(docker): exclude tags that transform to the current version

hktitof:
### Summary

when the registry only offers tags that the transformTags formula normalizes to the same version as the current one, both tags normalize to the same version but the candidate gate in getTagCandidates still accepts the rebuild

example: linuxserver sabnzbd image, current tag is 4.5.5-ls239 and the registry has 4.5.5-ls244, with a transform formula that strips the -lsNNN suffix

```yaml
labels:
  - 'wud.tag.include=^\d+\.\d+\.\d+-ls\d+$$'
  - 'wud.tag.transform=^(\d+\.\d+\.\d+)-ls\d+$$ => $$1'
```

the keep only greater semver gate uses isGreater which is semver gte, so an equal after transform tag passes it, and the semver sort compares transformed tags with the same gte, so transformed equal tags count as greater there too and the selected candidate ends up depending on the order the registry returns the tags, the same registry content can give result.tag = 4.5.5-ls239 or result.tag = 4.5.5-ls244

the stored candidate then contradicts the model itself, since updateAvailable in model/container.ts already decides updates by transformed tag inequality (localTag !== remoteTag), so WUD keeps a phantom next tag in its state while reporting no update

### Behavior / compatibility

- the gate now keeps a candidate only when its transformed value is different from the current transformed value AND greater semver, so getTagCandidates agrees with the updateAvailable rule and the result no longer flips with registry tag order
- strictly greater transforms and no transform flows are unchanged
- updateAvailable itself is untouched

### Test plan

added should not propose a tag that transforms to the same version as the current one in app/watchers/providers/docker/Docker.test.ts, it emulates the -ls suffix stripping transform and a gte comparison, then asserts the rebuild is not proposed

it fails without the source change (received 4.5.5-ls244 instead of 4.5.5-ls239) and passes with it

```
npx jest watchers/providers/docker/Docker.test.ts   # 58 passed
npx eslint watchers/providers/docker/Docker.ts watchers/providers/docker/Docker.test.ts   # 0 errors
```

relates to #938

### Checklist
- [x] odd behavior verified before the fix (pre-change probe with real modules, outcome depended on registry tag order)
- [x] unit test added that fails without the fix and passes with it
- [x] lint clean on changed files
- [x] no dependency / lockfile change

### DIFF
--- a/app/watchers/providers/docker/Docker.test.ts
+++ b/app/watchers/providers/docker/Docker.test.ts
@@ -989,6 +989,44 @@ describe('Docker Watcher', () => {
 
             expect(result).toEqual({ tag: '1.3' });
         });
+
+        test('should not propose a tag that transforms to the same version as the current one', async () => {
+            const container = {
+                includeTags: '^4.5.\\d+-ls\\d+$',
+                transformTags: '^(\\d+\\.\\d+\\.\\d+)-ls\\d+$ => $1',
+                image: {
+                    registry: { name: 'hub' },
+                    tag: { value: '4.5.5-ls239', semver: true },
+                    digest: { watch: false },
+                },
+            };
+            const mockRegistry = {
+                getTags: jest.fn().mockResolvedValue([
+                    '4.5.5-ls244', // rebuild of 4.5.5 - same version after transform
+                    '4.5.5-ls239', // current tag
+                ]),
+                shouldWatchDigest: jest.fn(() => false),
+            };
+            registry.getState.mockReturnValue({
+                registry: { hub: mockRegistry },
+            });
+
+            // Emulate the real transformer (strip the -lsNNN rebuild suffix)
+            mockTag.transform.mockImplementation((formula, tag) =>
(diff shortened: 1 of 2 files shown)


hktitof:
hey, about the red E2E check here: it is unrelated to this diff and it hits every PR run since yesterday evening, here is what i found

the failing scenarios are all the gitlab ones (gitlab_test container), they fail with "Request failed with status code 401" bcs gitlab.com's /jwt/auth endpoint started rejecting invalid basic credentials with a hard 401 for public projects, and our e2e compose falls back to WUD_REGISTRY_GITLAB_PRIVATE_TOKEN=dummy when the GITLAB_TOKEN secret is unset, so the registry client can't get a token and tag listing fails

lint/unit/build are all green on this PR and the change only touches the docker watcher, i opened #1165 with a small fix (optional gitlab token + anonymous pull for public projects, which the docs already promise), once it lands a rebase here should turn this check green

PR #1165, opened 2026-08-30
fix(gitlab): allow anonymous pull for public projects

hktitof:
### Summary

the gitlab scenarios in the e2e suite went red on every PR run since yesterday evening (401 on the gitlab_test container), so i dug into it and it is not flaky, gitlab changed the auth behavior

root cause: gitlab.com `/jwt/auth` endpoint now rejects invalid basic credentials with a hard 401 even when the project is public, it used to fall back to an anonymous token before. our e2e compose falls back to `WUD_REGISTRY_GITLAB_PRIVATE_TOKEN=dummy` when the `GITLAB_TOKEN` secret is unset, and the gitlab registry client always sent that bogus basic header, so the token request fails and no tags can be listed

repro with plain curl:

```
$ curl -s -o /dev/null -w '%{http_code}\n' -H "Authorization: Basic $(printf ':dummy' | base64)" "https://gitlab.com/jwt/auth?service=container_registry&scope=repository:gitlab-org/gitlab-runner:pull"
401

$ curl -s -o /dev/null -w '%{http_code}\n' "https://gitlab.com/jwt/auth?service=container_registry&scope=repository:gitlab-org/gitlab-runner:pull"
200
```

the gitlab docs page already advertises zero-config for public images, so this PR makes the code match that promise:

- `token` becomes optional in the schema (empty by default), public projects now need no configuration at all
- `authenticate()` only sends the basic header when username or token are actually configured, anonymous pulls still go through `/jwt/auth` to get the registry bearer token (the registry itself requires it, verified `tags/list` works fine with the anonymous bearer)
- e2e compose no longer falls back to a dummy token
- docs mark the token as optional

private registries keep working exactly as before, a configured token is still sent as basic auth

### Test plan

- [x] unit tests: anonymous flow sends no basic header, empty config validates to defaults (13/13 in Gitlab.test.ts)
- [x] full e2e suite green locally with this change, gitlab scenarios included


fmartinou:
Thanks!

Issue #1154, opened 2026-08-27
Authentication session store can overwrite newer container records in `wud.json`


NorskNoobing:
## Description

WUD appears to open the same LokiJS database file through two independent database instances:

1. The main WUD store for application and container state.
2. The `connect-loki` Express session store.

Because each instance loads and autosaves its own complete snapshot, a session-store write can overwrite container changes made by the main store.

## Environment

- WUD image: `getwud/wud:8.3.1`
- Persistent store: `/store/wud.json`
- Docker watcher with a daily cron scan
- OIDC and Basic authentication enabled
- `connect-loki`: 1.2.0

## Observed behavior

Two running containers currently use:

```text
ghcr.io/goauthentik/server:2026.5.6
```

WUD’s live API correctly reports:

```text
local:  2026.5.6
remote: 2026.8.0
```

At the same time, the persisted `containers` collection in `/store/wud.json` contains:

```text
local:  2025.12.1
remote: 2026.2
```

After WUD restarts, it displays the stale `2025.12.1` records.

WUD previously sent notifications for releases newer than `2025.12.1`, proving scheduled scans had detected them. Those results existed in the running process but were not reliably retained in the database file.

## Likely cause

The primary store creates a LokiJS instance in `dist/store/index.js`:

```js
const db = new Loki(`${configuration.path}/${configuration.file}`, {
    autosave: true,
});
```

The authentication code creates a second LokiJS instance through `connect-loki`, using the same path:

```js
store: new LokiStore({
    path: `${store.getConfiguration().path}/${store.getConfiguration().file}`,
    ttl: 604800,
}),
```

`connect-loki` creates its own autosaving LokiJS client:

```js
this.client = new Loki(this.storePath, {
    env: 'NODEJS',
    autosave: self.autosave,
    autosaveInterval: 5000
})
```

The two LokiJS instances do not appear to coordinate their in-memory state. When an authenticated request creates or touches a session, the session-store instance can save its older snapshot of the `containers` collection over the newer snapshot written by the watcher.

Frequent authenticated metrics requests made the issue especially visible, but ordinary OIDC sessions or authenticated API requests can also trigger it.

## Additional session issue

The session collection grew to more than 540,000 records.

`connect-loki` creates it using:

```js
self.collection = self.client.addCollection('Sessions', {
    indices: ['sid'],
    ttlInterval: self.ttl
})
```

[...]

PR #1171, opened 2026-09-01
Fix storage by reusing existing storage for loki

Tom-V:
## Summary

Replace `connect-loki` with a native Loki-backed Express session store that reuses WUD’s existing Loki database.

## Changes

- Encapsulate the Loki database and store configuration in a shared `Store` instance.
- Add `LokiSessionStore` implementing the required `express-session` store methods
- Reuse the existing `Sessions` collection when the database is already initialized.
- Persist sessions in WUD’s configured Loki database.
- Remove the `connect-loki` dependency.
- Handle expired sessions using:
  - The session cookie’s expiry date when available.
  - A seven-day TTL fallback otherwise.

## Why

The previous `connect-loki` integration opened its own Loki database connection. This could result in the session store using a different database instance from the rest of WUD, preventing reliable session persistence and database reuse.
Depending on what saved at what time, sessions or the main database could have been overridden by the other save.

The new implementation shares WUD’s initialized Loki database and `Sessions` collection.

Fixes 
- #1154

### DIFF
--- a/app/api/LokiSessionStore.test.ts
+++ b/app/api/LokiSessionStore.test.ts
@@ -1,0 +1,154 @@
+import Loki from 'lokijs';
+import session from 'express-session';
+import LokiSessionStore from './LokiSessionStore';
+
+describe('LokiSessionStore', () => {
+    test('should reuse the existing sessions collection on the provided db', () => {
+        const db = new Loki('test.db');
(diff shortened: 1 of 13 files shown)


fmartinou:
Thanks

PR #1209, opened 2026-09-09
fix(docker): read the container id from Actor.ID, not the legacy flat id

piorkov:
Docker 29.x no longer sends the flat `id` / `status` fields on the events stream; the container id is only in `Actor.ID`. `containerId` was therefore always undefined, so onDockerEvent asked the daemon for `/containers/undefined/json` on every start/die and never refreshed a container's status. The exception is caught, so nothing was reported.

Verified on Docker 29.8.0: with the fix, "Status changed from running to exited" and back is logged again; without it, that line never appears.

The `?? dockerEvent.id` fallback keeps older daemons working.

Fix for https://github.com/getwud/wud/issues/1208

### DIFF
--- a/app/watchers/providers/docker/Docker.ts
+++ b/app/watchers/providers/docker/Docker.ts
@@ -447,7 +447,11 @@ export class Docker extends Watcher {
             return;
         }
         const action = dockerEvent.Action;
-        const containerId = dockerEvent.id;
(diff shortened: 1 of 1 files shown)


fmartinou:
Thank you for your contribution; I included it in this other PR
https://github.com/getwud/wud/pull/1214

Issue #484, opened 2024-11-13
Error self-updating wud


carlosjfcasero:
Every time a new wud version is published, my local wud crashes. Here are the logs:

```wud  | 10:57:26.642  INFO whats-up-docker/trigger.docker.1: Image getwud/wud:latest pulled with success (container=local_wud)
wud  | 10:57:26.644  INFO whats-up-docker/trigger.docker.1: Stop container wud with id 39b17d476afb1183356c0c2099d025961ad549ec2cba815968c533c23f774899 (container=local_wud)
wud  | 10:57:32.534  WARN whats-up-docker/watcher.docker.local: Error when processing some containers (Cannot read properties of undefined (reading 'tag'))
wud  | 10:57:32.534  INFO whats-up-docker/watcher.docker.local: Cron finished (0 containers watched, 0 errors, 0 available updates)
wud  | 10:57:36.508  INFO whats-up-docker/watcher.docker.local: Cron finished (25 containers watched, 0 errors, 1 available updates)
wud exited with code 137
```
Since it stops the container before creating the new one, it's not able to continue and wud stops. I have to enter manually and run `docker compose up` again. I could also disable auto updates for wud image but that's precisely the purpose of wud.

Thilas:
Same issue here.

@fmartinou, is self-update supposed to be supported? The documentation (https://getwud.github.io/wud/#/configuration/?id=complete-example) suggests it is. But as mentioned by @carlosjfcasero, once the container is stopped, it is never started again.

sahps:
Having the same issue.

Config file is:
```
services:
  wud:
    image: getwud/wud:8.1.0
    labels:
      wud.watch: true
      wud.trigger.include: smtp.notify,dockercompose.wud
      wud.tag.include: ^\d+\.\d+\.\d\d?\d?\d?$$
    container_name: wud
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      - /data/docker/library:/wud/mnt/local
    ports:
      - 9080:3000
    environment:
      # wud authentication
      WUD_AUTH_BASIC_MY_USER: ######
      WUD_AUTH_BASIC_MY_HASH: ######
      # smtp notifications
      WUD_TRIGGER_SMTP_NOTIFY_HOST: #.#.#.#
      WUD_TRIGGER_SMTP_NOTIFY_PORT: 25
      WUD_TRIGGER_SMTP_NOTIFY_FROM: wud@example.com
[...]

PR #1257, opened 2026-09-15
🐛 [TRIGGER] Fix docker trigger leaving WUD stopped when it updates its own container (fixes #1241)

### LINKED ISSUE #1241: WUD fails to restart when it updates itself
Running on Debian 13 in a Proxmox VM.
In previous versions (8.x) WUD was able to restart after it updates itself, however this morning's the container exitted, completed the update but never restarted.  I was able to start it manually with no issues.  Below is my docker-compose (not sure where the logs are stored and are recoverable for previous images of the container):
services:
  whatsupdocker:
    image: getwud/wud
    container_name: wud
    hostname: wud
    restart: unless-stopped
    networks:
      - proxy
    env_file:
      - .env
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      - ./store:/store
    ports:
      - 3000:3000
    labels:
      wud.watch.digest: "true"

networks:
  proxy:
    external: true

w1mm3r:
## Problem

When the docker trigger updates the container WUD itself runs in, `trigger()` performs the swap in-process:

```
pull → clone spec → stop → remove → create → start
                     ↑
                     the process doing the work is the one being stopped
```

It never gets past `stop`. WUD exits 137, the replacement is never created, and `restart: always` does not help because a deliberate `docker stop` does not trigger a restart policy. The container simply stays down.

From a real instance, 8.3.1 — the update was detected and then nothing happened for seven days:

```
14:00:07  Pruning previous tags (container=local_wud)
14:00:07  Pull image getwud/wud:latest (container=local_wud)
14:00:33  Image getwud/wud:latest pulled with success (container=local_wud)
14:00:33  Stop container wud with id 576f882a... (container=local_wud)
          <exited with code 137>
```

Reported in #1241 (open) and #484 (closed as not planned, same log signature). The worst part is that it is silent: WUD is the thing that tells you about updates, so when it dies nothing tells you it stopped watching.

## What this changes

**1. It no longer fails silently (default behaviour).**

The trigger now detects that the target is WUD's own container and refuses, with an error that says what to do:

```
Refusing to update the container WUD runs in: stopping it would abort the update
and leave the container down. Enable it with
WUD_TRIGGER_DOCKER_{trigger_name}_SELFUPDATE=true, or exclude WUD from this
trigger with the wud.trigger.exclude label and update it externally.
```

Nothing is stopped or removed. This is the part that fixes #1241 for everyone, and it is behaviour-compatible: today the update does not work anyway, it just breaks the container while failing.

**2. It can actually do it, opt-in.**

`WUD_TRIGGER_DOCKER_{trigger_name}_SELFUPDATE=true` delegates the swap to a short-lived helper container, because the work has to outlive the stop. The sequence is the one `portainer-updater` uses in `dockerstandalone/update.go`:

1. create the replacement under a temporary name, before touching anything
2. stop the old container
3. start the replacement
4. wait for it to become healthy (`WUD_TRIGGER_DOCKER_{trigger_name}_SELFUPDATETIMEOUT`, default 120s)
5. success → remove the old container, rename the replacement
6. failure → remove the replacement, restart the old container

The helper runs **the image WUD is running right now**, not the one being installed. That avoids a chicken-and-egg problem: the code performing the swap is always known to contain this feature, so it works on the very first release that ships it. It mounts the same socket as the watcher.

It is named `wud-self-update` and deliberately *not* auto-removed. If a swap fails, WUD is not running to record why, so the helper's logs are the only account of what happened; the next self-update removes the previous helper first, so at most one is ever left behind.

Watchtower solves the same problem by renaming itself and letting the old and new containers overlap. That is not available here — WUD publishes a port, so two instances cannot run at once. Hence the helper.

## Detecting WUD's own container

`/proc/self/mountinfo` is the primary source: Docker bind-mounts `/etc/hostname`, `/etc/hosts` and `/etc/resolv.conf` from `<data-root>/containers/<id>/`, so the full id is there regardless of data-root.

`HOSTNAME` is only used as a last resort, and only when it still resolves to a live container whose id it prefixes. That verification matters: `cloneContainer` copies `Config.Hostname` onto replacements, so **every container WUD has already recreated carries the previous container's id as its hostname**. On a live instance:

```
radarr     id=fbd85cb413de  hostname=edbd3f68039d   <- id of the container WUD replaced
prowlarr   id=37ac67058824  hostname=992e8969b910   <- same
wud        id=d0478132ffc6  hostname=d0478132ffc6   <- created by compose, still correct
plex       id=401b6145724a  hostname=plexserver     <- explicit hostname: in compose
```

Trusting `HOSTNAME` would mean mistaking another container for ourselves. `WUD_CONTAINER_ID` is available as an explicit override for runtimes that expose neither.

In the self-update path the replacement's `Hostname` is dropped from the clone so Docker assigns a fresh one, rather than inheriting this container's id.

## Tests

`app/triggers/providers/docker/self.test.ts` (33 tests) covers id resolution (all three strategies, plus the stale-hostname, wrong-container and non-hex cases), the health gate (healthy / unhealthy / died / no healthcheck / inspect failure / timeout), and the swap, asserting exact call order for the happy path and for each rollback path.

`app/triggers/providers/docker/Docker.test.ts` adds 4 tests: the default refusal (asserting nothing is stopped or removed), the helper delegation (asserting the helper spec, the payload, and that this process stops nothing), the remote-host rejection, and that a normal container still follows the existing path.

`cd app && npm test` (1325 tests) and `npm run lint` pass. The branch is rebased on current `main` (`7d639d3`), including the `updateAvailable` early return added there. Docs build cleanly and pass `lint:md` and `lint:spell`. `changelog/next.md` updated.

### Verified against a real daemon

Unit tests cannot prove the part that matters — that the swap survives the process being stopped — so both paths were also run against a live Docker daemon, using a throwaway container built from this branch:

```
1. HAPPY PATH: helper replaces a running, port-publishing container
   Self-update: created replacement container wud-e2e-target-wud-self-update-...
   Self-update: stopped wud-e2e-target
   Self-update: started wud-e2e-target-wud-self-update-...
   Self-update: removed the previous wud-e2e-target
   Self-update: renamed the replacement to wud-e2e-target
   PASS  container replaced
   PASS  old container removed
   PASS  replacement renamed to 'wud-e2e-target'
   PASS  replacement is running
   PASS  replacement re-bound host port 13000
   PASS  no temporary containers left behind

2. ROLLBACK: replacement cannot start -> old container restored
   Self-update failed (unable to start the replacement (... "/nonexistent-binary" ...)); rolling back
   Self-update: restarted wud-e2e-target
   PASS  original container was restarted after the failure
   PASS  original kept its name
   PASS  failed replacement was cleaned up

RESULT: 9 passed, 0 failed
```

The port assertion is the one that matters for the design: the replacement binds the same host port the old container held, which is only possible because the old one is stopped first — and is exactly why watchtower's rename-and-overlap approach cannot be reused here.

The health gate took ~31s in that run, which is the image's 30s healthcheck interval; the 120s default leaves room for it.

## Limitations

- Requires the watcher to use a Docker socket; rejected with a clear error for `WUD_WATCHER_{watcher_name}_HOST`.
- The helper needs to be able to create containers on that socket, which the docker trigger already requires.
- Swarm/Kubernetes are unaffected — this is the standalone docker trigger only.


fmartinou:
Looks great!

Thanks you very much!

aam1972:
>     * Swarm/Kubernetes are unaffected — this is the standalone docker trigger only.

Hello, will this work with docker compose?

thanks, Andrea

fmartinou:
Hi @aam1972 ,

Yes! It works if your WUD container is managed by Docker Compose:

With the Docker trigger (WUD_TRIGGER_DOCKER_{trigger_name}_SELFUPDATE=true), the helper container swaps the container via the Docker socket (it does not edit your docker-compose.yml file).

With the Docker Compose trigger (WUD_TRIGGER_DOCKERCOMPOSE_{trigger_name}_SELFUPDATE=true), it updates the image tag in your compose file first, then delegates the container swap to the helper.

PR #1277, opened 2026-09-21
🐛 [WATCHER] Fix container temporary name persistence on rename (fixes #770)

### LINKED ISSUE #770: WUD gets stuck with initial temporary container name
When I manually update a container, I change the version number in the compose file and redeploy it ("Update the stack" in Portainer). After downloading the new image, it kind of closes the old container and starts the new one simultaneously. I guess they aren't actually running at the same time, but there's a brief moment when the list of containers has both the old and the new version in various states of existance. Since they would have the same name, the new one gets prefixed with some hash-like string. When the old container finally gets removed, the new container gets renamed.

WUD picks up the initial name and keeps it for as long as the container is running. The name only gets updated if I restart that container (basically killing it, and creating a new one).

Could this be fixed so that WUD also gets the updated container name?

Here's Portainer in the process of updating a container:
Image

This is the result when the update is done:
Image

But WUD gets stuck with the initial temporary name:
Image

fmartinou:
## Problem
When tools (such as Docker Compose, Portainer, Coolify, or Kamal) create containers with temporary names and subsequently rename them to their final target name, WUD remains stuck displaying the initial temporary container name:
1. `listenDockerEvents()` filtered container events without including the `'rename'` event.
2. `onDockerEvent()` handled container events by only synchronizing container status (`containerFound.status = newStatus`), never updating `containerFound.name`.
3. Scheduled polling in `addImageDetailsToContainer()` found the existing cached container in the store and returned it without checking whether the container name had changed.

## Solution
1. Added `'rename'` to the container event filter list in `Docker.listenDockerEvents()`.
2. Updated `Docker.onDockerEvent()` to inspect the container and synchronize `containerFound.name` with the store when a rename or update occurs.
3. Updated `addImageDetailsToContainer()` to detect name changes for cached store containers, update `containerInStore.name`, and persist the change with `storeContainer.updateContainer()`.
4. Enhanced `getContainerName()` to safely extract the container name from `container.Names`, `container.Name`, and `container.name`, stripping leading slashes.
5. Added comprehensive Jest unit tests covering rename event processing, polling name reconciliation, and container name parsing.
6. Updated changelog in `website/docs/changelog/next.md`.

## Validation
- Ran full backend test suite (`cd app && npm test`): 111 suites passed, 1397 tests passed.
- Ran backend linter (`cd app && npm run lint`): 0 errors.
- Ran website docs linter (`cd website && npm run lint:docs`): 0 errors.
- Ran website build (`cd website && npm run build`): build successful.

Fixes #770


alexdelprete:
I can't see this in 9.1.0 release notes. Has it been included?

fmartinou:
`9.1.0` was released before this fix, so it will be included in the next release.

Current unreleased changes are listed here:
https://getwud.app/docs/next/changelog/next

alexdelprete:
> `9.1.0` was released before this fix, so it will be included in the next release.

I saw this merged 3 days ago and release was 2 days ago. That's why I asked. :)

Ok, will wait for next release to test it. Thanks.

